PR #168, opened 2025-05-16
Feature/Allow Transcoding of Proxied Channel Restream

JoKneeMo:
This feature allows you to override the codec used when proxy is enabled.

My provider uses AAC for the audio stream which required transcoding on Plex's desktop client.
If the audio codec is `libopus`, no transcoding is required and Direct Play and Direct Stream can be used.
I also prefer the DVR to be encoded with h265 instead of the provider's default h264.

When a value is set, it will transcode to that codec.
If the value is empty, it will copy the original stream's codec as it always has.

The values can be set from the `config/env` file, docker config, or the settings UI.
The executed value will go in order from environment → settings UI → `copy`.

New optional variables:
- PROXY_FFMPEG_CODEC_VIDEO
- PROXY_FFMPEG_CODEC_AUDIO
- PROXY_FFMPEG_CODEC_SUBTITLES

From the settings UI, the available codecs are automatically loaded into the dropdown select field.
These values are generated from the output of `ffmpeg -encoders`  executed and cached when the page loads, so that it's not called frequently.

### Settings Screenshot
image

In the screenshot above:
- Video codec: `libx265` has been selected from the settings UI.
- Audio codec: The environment file contains `PROXY_FFMPEG_CODEC_AUDIO=libopus`.
  You can see that it will disable input from the settings panel when an environment variable is used.
  The value will be shown in the UI.
- Subtitles codec: No value is set in the settings UI, nor an environment variable, so it will default to `copy`.

### ShowConfig Command
Also included is a debugging command to show config variables from the terminal.
Example:
```bash
# php artisan config:show proxy
proxy
array:5 [
  "url_override" => null
  "ffmpeg_additional_args" => ""
  "ffmpeg_codec_video" => ""
  "ffmpeg_codec_audio" => "libopus"
  "ffmpeg_codec_subtitles" => ""
] // app/Console/Commands/ShowConfig.php:23
```


sparkison:
This looks great, thanks so much for the help!! 🙏

PR #169, opened 2025-05-18
Option to use jellyfin-ffmpeg

Redth:
The current `ffmpeg` in alpine is 6.x and has issues handling some streams.  Jellyfin ships a `jellyfin-ffmpeg` package based on 7.x which seems to do better for these streams.

This installs the `jellyfin-ffmpeg` package in the Dockerfile, and adds an option to set the environment variable `FFMPEG_USE_JELLYFIN=true` which will cause that ffmpeg to be set as the `FFMPEG_PATH` environment variable which is then used in the streaming code later.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN apk update && apk --no-cache add \
     nodejs \
     npm \
     ffmpeg \
+    jellyfin-ffmpeg \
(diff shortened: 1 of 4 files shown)


sparkison:
Thank you for the PR!

PR #171, opened 2025-05-20
Update PlaylistUrlService.php

drondeseries:
Description of Changes:

1.  Added null checks to the getUrls method in PlaylistUrlService to prevent "Trying to access array offset on null" errors.

2.  Now, before accessing ['key'] on each short URL type (m3u, hdhr, epg, epg_zip), the code verifies that the entry exists and is not null.

3. This improves the stability of the URL generation and prevents view exceptions when certain short URLs are missing.


sparkison:
Doh! Thank you for catching this

PR #187, opened 2025-06-01
Feature/group sort order

Redth:
Adds a sort order field for groups and sorts the playlist output by groups first, then the existing channel sort.


sparkison:
Looks good! Might make the sort column a double as I've had request on the channels to allow decimals for more precision. Other than that, I'll get this added to the next dev build, and ultimately into the next release!

thanks for the assist 😉

PR #235, opened 2025-06-14
Add timeshift support to MPEG-TS

djrarky:
Adding support for timeshift. 

to-note: may need to extend default ffprobe timeout.

### DIFF
--- a/app/Http/Controllers/StreamController.php
+++ b/app/Http/Controllers/StreamController.php
@@ -47,6 +47,26 @@ class StreamController extends Controller
         $sourceChannel = $channel;
         $streams = collect([$channel])->concat($channel->failoverChannels);
 
+        /* ── Timeshift parameters, if the request asked for them ───────────── */
+        $utcPresent = $request->filled('utc');
+        if ($utcPresent) {
+            $utc    = (int) $request->query('utc');              // programme start
+            $lutc   = (int) ($request->query('lutc') ?? time()); // “live” (default = now)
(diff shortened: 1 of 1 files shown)


djrarky:
I explored adding this to HLS streaming as mentioned in the discussion https://github.com/sparkison/m3u-editor/discussions/225 but realise how it now works. 
Probs best to leave support missing for HLS, might be worth removing the catch-up icon from tivimate etc. for HLS configs too.

djrarky:
My code is garbage. 
catch-up stuff is 2 hours out (ie I play footage from 9am, but it plays from 7) and the redirect to MPEG-TS doesn't work. 

Just fyi haha, I'll leave it up tho as it's a good step forward.

sparkison:
awesome! This is a great start! The whole 2hr out thing, you're saying it's off by 2hr? Does changing the `TZ` variable fix it? Just curious before I grab it to merge. Unfortunately I don't have any streams with catchup so I'll be leaning on you to do some testing once implemented 😉

sparkison:
...follow up! I see you have "utc" hardcoded in the timestamp logic. Just curious if we used the container TZ if that would fix the off by 2hr issue?

djrarky:
I need to look into it some more, definitely not ready to be merged.

That's a good shout, possibly. I struggle with TZ as the server, my parents and I are all in different timezones. I'll spend some more time on this.

Thanks mate.

PR #239, opened 2025-06-16
corrected naming of webhook routes

djrarky:
Corrected the naming of webhook routes.
POST endpoint uses webhook.test.post, GET endpoint uses webhook.test.get


sparkison:
haha! Good catch 😄

djrarky:
Haha that one bugged me

PR #271, opened 2025-07-18
Dev

drondeseries:
fixed merge same id where only one channel was being created.

### DIFF
--- a/app/Filament/Resources/ChannelResource.php
+++ b/app/Filament/Resources/ChannelResource.php
@@ -29,6 +29,7 @@ use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Str;
 
 class ChannelResource extends Resource
 {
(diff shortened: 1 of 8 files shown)


drondeseries:
Here's a summary of the changes:

Failover Playlist Selection: The 'Merge Same ID' action modal now includes a repeater field that allows you to select one or more playlists to be used as failover sources. You can add, remove, and reorder the failover playlists.

Resolution-based Ordering: A new 'Order by Resolution' toggle has been added to the modal. When enabled, the failover channels will be sorted by resolution, with the highest-resolution channel being prioritized first. This provides a more intelligent failover mechanism, but it may take longer to process.

Preferred Playlist in Failover: The preferred playlist is now automatically included in the list of failover playlists. This ensures that all channels with the same stream ID from the preferred playlist are available as failover options.

Job Modifications: The MergeChannels job has been updated to handle the new parameters. It now accepts a collection of failover playlists and a boolean flag to indicate whether to order by resolution. The job has also been updated to correctly handle different data formats for the playlist data.

These changes provide a more flexible and powerful way to manage channel failover, giving you more control over your merged playlists.

sparkison:
This doesn't work as it fetches ALL of the channels ahead of running the job and passes them as an array. May work fine for small lists, or only a few playlists. But in my case, it fails and the job doesn't even execute, instead I simply get an out of memory error. Try to fetch the latest `experimental` branch first and work from that. Going to close this one though as it doesn't work.

PR #292, opened 2025-08-13
Implement timeshift function

djrarky:


djrarky:
Hey, so this is added - it does however require either the providers timezone or timeshift in hours so that the timestring is accurately calculated to relevant time on the providers server. 

I thought this could be added against the playlist but I'm not quite sure how to do that. This works well!

sparkison:
> Hey, so this is added - it does however require either the providers timezone or timeshift in hours so that the timestring is accurately calculated to relevant time on the providers server.

woohoo! 🎉  Thanks for your work on this! I'll take a look and see if I can finish it out

PR #313, opened 2025-08-31
Allow short URLs to forward HDHomeRun subpaths

djrarky:
Fixes https://github.com/sparkison/m3u-editor/issues/310


sparkison:
looks good!

PR #324, opened 2025-09-06
Custom playlist - View channels within groups from the group page

giejay:
Sort channels within groups by drag/drop and alphabetically

This will add a "view" icon to the group page within a custom playlist. The view icon pops up a modal in which the channels can be sorted, either by dragging or by a button: alphabetically. 

In the end, it would be nice to not use the "sort" column of the channel, but create a pivot table so you can sort the same channels differently within multiple groups. But I don't know exactly how the whole export to m3u works, so didn't touch that. 

Disclaimer: I had help from Copilot since I'm not familiar with Lavarel framework:) So, if I coded any anti patterns, I'm sorry, please let me know.

### DIFF
--- a/app/Filament/Resources/ChannelResource.php
+++ b/app/Filament/Resources/ChannelResource.php
@@ -98,7 +98,8 @@ class ChannelResource extends Resource
             ->columns(self::getTableColumns(showGroup: !$relationId, showPlaylist: !$relationId))
             ->filters(self::getTableFilters(showPlaylist: !$relationId))
             ->actions(self::getTableActions(), position: Tables\Enums\ActionsPosition::BeforeCells)
-            ->bulkActions(self::getTableBulkActions());
+            ->bulkActions(self::getTableBulkActions())
+            ->reorderable('sort'); // Enable drag-and-drop sorting
     }
 
     public static function getTableColumns($showGroup = true, $showPlaylist = true): array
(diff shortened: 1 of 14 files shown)


app/Filament/Resources/ChannelResource.php
```diff
@@ -133,14 +134,9 @@ public static function getTableColumns($showGroup = true, $showPlaylist = true):
...
                 ->toggleable(),
-            Tables\Columns\TextInputColumn::make('sort')
+            Tables\Columns\TextColumn::make('sort')
                 ->label('Sort Order')
-                ->rules(['min:0'])
-                ->type('number')
-                ->placeholder('Sort Order')
                 ->sortable()
-                ->tooltip(fn($record) => !$record->is_custom && $record->playlist?->auto_sort ? 'Playlist auto-sort enabled; disable to change' : 'Channel sort order')
-                ->disabled(fn($record) => !$record->is_custom && $record->playlist?->auto_sort)
```
giejay, line 143: Might need to make the reorderable('sort') dependant on the same logic...

sparkison:
it is a little anti-pattern with the additional JS lib and controller to handle the sorting. I've implemented something more inline with the framework that should accomplish the same thing. This is on `experimental` now 😉 

sort

giejay:
Understood! Thanks for adding it yourself, I will try it now:) One question, would it be possible to add the alphabetical sorting to the bulk action dropdown?

giejay:
I tried it, and it doesnt work correctly Im afraid. It only shows the latest added group and it hides all the other channels. So with 1 group it works, but after adding another group, only that group is shown, both in the Live channels tab and in the groups tab.

PR #328, opened 2025-09-11
Implemented playlist syncing

djrarky:
Added playlist syncing using a parent–child playlist system.

* Child playlists must use the same provider, but with different login details (useful when each login only allows one stream).
  - I’ve deliberately allowed the provider URL to vary, since some providers use different hostnames or CDNs for the same content. This does not cover the edge case of switching to a different provider and expecting sync to work.
  - To make this clear, syncing only happens when duplicating a playlist, via the “Duplicate with sync” option.
* Deduplicated parent–child items (live channels, series, VOD, categories, groups, etc.) in the GUI and added a Playlist column to these tables.
  - The column shows the parent playlist name plus the number of child playlists. Hovering over it reveals the full list of playlists the item belongs to.
  - Custom playlists also display the item’s source playlist (the parent), and you can change this in-line when the item is part of a parent–child group.

Any changes made to a parent playlist are automatically passed down to its children:

* instantly, as changes happen, and
* in bulk when syncing with the upstream provider.

Also added support for parent–child groups when adding playlists to custom collections, with a new modal for this purpose. The option to select a source playlist only appears when it’s relevant.

The “merge same ID” feature now works with parent–child playlist groups for those using failover.


djrarky:
I hope this isn't garbage mate. Tried my best with copilot.

sparkison:
hey! Great work!

that said, I likely won't merge this due to the framework migration from Filament v3 to Filament v4 on `experimental`, and the fact all branches will eventually adopt what's on experimental now, so this would cause a bit of back-tracking.

It would be a lot of code to try and copy over to the updated app since the folder structure has changed unfortunately.

However, I think I might have a simple solution for this! Is the goal to just keep the playlists in sync? No other reason for it? E.g. so you can have one "parent" that all "children" sync to? So that you can share the other lists with family, etc. and they receive the same changes that are applied to the parent?

Just trying to make sure I'm following! If this is the case, thinking I can create a sort of "shadow" playlist. It would be an empty playlist that would just use the primary/master/parent playlist as it's "source" and adjust the URLs accordingly. This would reduce the data, so it's not duplicated, while allowing "synced" lists 🎉 

Let me know what ya think 😉

djrarky:
Hey mate, 

Yeah that makes tons of sense, better way being a shadow playlist. My only thought was, what if you wish to use a different provider url, I tried to account for that. 

The handling of custom playlists with this functionality I was pretty pleased with tho - hopefully you can rip that!

Basically yeah, just to keep the playlists in sync. My nan can have a provider, mum and then me - when making mass changes I currently delete the playlists, make the changes - then duplicate.

Thank you :)

sparkison:
My initial take of this feature is up on `experimental` now! Try it out and let me know how it goes! Or if I need to tweak anything for your use case 😉 hoping I got it though 🤞

alias

djrarky:
Appreciate it a ton - thanks mate. 

Is there any chance you could allow a differing proxy config on the alias?

sparkison:
> Appreciate it a ton - thanks mate.
> 
> Is there any chance you could allow a differing proxy config on the alias?

no problem! And done! 😉 https://github.com/sparkison/m3u-editor/commit/783ee8c21fd91ab00f743da49642bfaad9beec73

djrarky:
Hey mate, I have another request if possible!

Can we show playlist expiration for the alias?

PR #393, opened 2025-10-06
Add options to Channels api

wschwa:
This change enables the "stream_id" to be modified via the new Channels API, using the same method as already implemented for name/logo/url.

Not a developer, so hopefully this change makes sense.

EDIT:
Additionally added the ability to enable or disable a channel via the same Channels API.


sparkison:
Looks good! Merged 😉

PR #448, opened 2025-11-09
Fix hostnames for postgres and redis in compose reference

bmwagner18:
The hostnames of postgres and redis were incorrect in the "external-all" variant of the docker-compose example. They needed the "m3u-" prefix to match the defined hostnames from their respective container within the compose file.


sparkison (approved):
This looks good, thanks for the assist! 🙏

PR #453, opened 2025-11-10
feat: Improve EPG channel mapping accuracy

Serph91P:
- Add three-tier matching approach (exact channel_id, exact name, then fuzzy)
- Implement stricter similarity thresholds to prevent false positives
- Add exact normalized match step (ignoring spaces, dashes, underscores)
- Require minimum 60% similarity for fuzzy matches
- Improve candidate filtering and validation
- Add detailed documentation for EPG mapping improvements


sparkison:
Looks good to me! Merging now! Lemme know how it goes 😉

sparkison:
...p.s. merged this into `dev` and `experimental`

PR #472, opened 2025-11-18
Optimize SimilaritySearchService by removing redundant COUNT query before cursor iteration

UnderwaterOverground:
**Summary**
This PR optimizes SimilaritySearchService::findMatchingEpgChannel() by eliminating a redundant COUNT(*) query on the EPG channels builder before streaming results with cursor().

**Details**
Removed the if ($epgChannels->count() === 0) pre-check, which executed a full COUNT(*) with all fuzzy LIKE and JSON conditions.
Introduced a $hasEpgChannels flag that is set while iterating foreach ($epgChannels->cursor() as $epgChannel).
After iteration, if no rows were seen, the method logs the same debug message and returns null, preserving existing behavior.

**Why**
count() + cursor() was causing the heavy EPG search query to be executed twice per call.
This method can be invoked tens of thousands of times (e.g., during bulk EPG matching), so removing the redundant query significantly reduces DB load and latency.
The change maintains identical functional behavior and logging while improving performance and memory efficiency.

### DIFF
--- a/app/Services/SimilaritySearchService.php
+++ b/app/Services/SimilaritySearchService.php
@@ -186,19 +186,13 @@ class SimilaritySearchService
         $bestMatch = null;
         $bestEpgForEmbedding = null;
         $candidates = []; // Store all candidates with scores for better decision making
-
-        if ($epgChannels->count() === 0) {
-            if ($debug) {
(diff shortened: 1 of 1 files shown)


sparkison:
Merged, thank you! 🙏

PR #476, opened 2025-11-19
Updated to respect proxy settings for logo

UnderwaterOverground:
Previously the EPG Generator was using the stream proxy setting flag for deciding whether to proxy logos, rather than using playlists' logo proxy setting, which is more appropriate.


sparkison:
Great catch! Thank you 🙏

PR #481, opened 2025-11-20
Gitignore DS_Store

UnderwaterOverground:
Convenience for OSX devs, ignore the attributes file OSX sprinkles everywhere.


sparkison:
Can't believe I missed this one this whole time! 🤦

PR #486, opened 2025-11-23
EPG playlist view respects configured sort order

UnderwaterOverground:
This one needs a quick visual test/review before merge.

The core issue was that the EPG view wasn't actually reflecting the sort order promised elsewhere (defined under the 'Live Channels' tab as below): 

`NOTE: Playlist channel output order is based on: 1 Sort order, 2 Channel no. and 3 Channel title - in that order. You can edit your Playlist output to auto sort as well, which will define the sort order based on the playlist order.`

Original issue: playlist/EPG API responses keyed channels by channel_no, so JSON became an object and JS re-ordered keys numerically, making the guide appear sorted by channel number instead of the documented DB sort order.

Specifically-- 

app/Http/Controllers/Api/EpgApiController.php (lines 343-358) builds $playlistChannelData with the playlist’s sort-aware SQL order, but it stores each channel in an associative array keyed by $channelNo. When the controller later returns 'channels' => $channels (app/Http/Controllers/Api/EpgApiController.php (lines 375-523)), PHP encodes that structure as a JSON object whose keys are integers. JavaScript enumerates integer-like object keys in ascending numeric order, so the UI receives the right payload but iterating over it reorders everything by channelNo, defeating the “sort order, channel number, channel title” promise. This is why the guide appears to respect channel numbers first even though the SQL query is already sorted correctly.

EpgGenerateController still orders its cursor by sort → channel → title (app/Http/Controllers/EpgGenerateController.php (lines 90-108)), so the XML export honors the intended priority; the mismatch only happens in the API response consumed by the guide.

Fix summary: add sequential sort_index metadata when building channel payloads, update the Alpine viewer to keep a channelOrder array sorted by that index, and drive counts/virtual scroll from it so UI rows follow the intended order even with numeric IDs.

### DIFF
--- a/app/Http/Controllers/Api/EpgApiController.php
+++ b/app/Http/Controllers/Api/EpgApiController.php
@@ -35,6 +35,7 @@ class EpgApiController extends Controller
         // Pagination parameters
         $page = (int) $request->get('page', 1);
         $perPage = (int) $request->get('per_page', 50);
+        $offset = max(0, ($page - 1) * $perPage);
         $search = $request->get('search', null);
 
         // Get parsed date range
@@ -71,7 +72,7 @@ class EpgApiController extends Controller
                     });
                 })
                 ->limit($perPage)
-                ->offset(($page - 1) * $perPage)
+                ->offset($offset)
                 ->get();
 
             // Get the channel IDs from database records to fetch cache data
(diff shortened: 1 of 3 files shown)


sparkison:
Looks good!

PR #502, opened 2025-11-30
Update PlaylistService.php

Aldione:
- Added direct authentication support for PlaylistAlias using username/password pairs.
- Fixed missing return when authenticating via alias credentials.
- Improved authenticate() logic ordering to avoid false-negatives.
- Ensured consistent return structure: [playlist, authMethod, username, password].
- Prevented array/object mismatches by standardizing auth handling.
- Added small cleanup and safer model loading for alias-based authentication.

### DIFF
--- a/app/Services/PlaylistService.php
+++ b/app/Services/PlaylistService.php
@@ -382,6 +382,21 @@ class PlaylistService
             }
         }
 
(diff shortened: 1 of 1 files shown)


sparkison:
Looks good! Nice catch 😉

PR #503, opened 2025-11-30
Update PlaylistAlias.php

Aldione:
- Added getAuthObjectAttribute() to consistently expose PlaylistAlias credentials as an object.
- Ensures username/password behave like standard objects across the entire application.
- Improves compatibility with existing code that expects object-based authentication data.
- Prepares PlaylistAlias for direct authentication support in the Xtream API layer.


sparkison:
Merging, thanks for the assist! 🙏

PR #514, opened 2025-12-03
Fix/series categories not saving

Serph91P:
This pull request introduces a few targeted improvements to playlist category handling and import job processing. The changes focus on correctly mapping category sources, improving import job tracking, and updating status reporting.

**Category Source Handling:**

* Updated the category mapping in the playlist form dehydration logic to use `SourceCategory` instead of `Category`, ensuring that category names are sourced from the correct model during form processing.

**Import Job Processing Enhancements:**

* Added an `importSeries` boolean parameter to the `processChannelCollection` method in `ProcessM3uImport.php`, providing more control over series import behavior during batch processing.
* Set the `progress` field to `100` in the completion handler of `ProcessM3uImportSeriesComplete.php`, ensuring that playlist import status reflects full completion when series sync is done.


sparkison:
can't believe I missed this one! 🫣 thanks again for the help! 🙏

PR #515, opened 2025-12-04
feat: Add separate VOD sync progress tracking and chunked processing

Serph91P:
- Add vod_progress column to playlists table for independent VOD tracking
- Add VOD Sync column to playlist table view with real-time progress
- Refactor VOD processing into chunked jobs (100 channels per chunk)
  - ProcessVodChannels: Dispatches job chain for bulk processing
  - ProcessVodChannelsChunk: Processes individual chunks with progress updates
  - ProcessVodChannelsComplete: Finalizes VOD sync and sets completion status
- Update status logic: Status stays 'processing' until ALL syncs complete
  - Channel, Series, and VOD must all finish before status becomes 'completed'
  - Each completion job checks if other syncs are still running
- Make manual 'Fetch VOD Metadata' action consistent with Series action
  - Sets status to 'processing' when started
  - Sets status to 'completed' when finished (if nothing else running)
- Add VOD progress to ImportProgress widget in playlist edit view
- Dispatch VOD jobs on 'import' queue for sequential processing
- Add vod_progress reset to all relevant reset actions

This prevents timeouts with large VOD libraries (16k+ channels) and provides clear visibility into what's currently syncing.

### DIFF
--- a/app/Console/Commands/ResetSyncProcess.php
+++ b/app/Console/Commands/ResetSyncProcess.php
@@ -59,6 +59,7 @@ class ResetSyncProcess extends Command
                     'errors' => null,
                     'progress' => 0,
                     'series_progress' => 0,
+                    'vod_progress' => 0,
                 ]);
             }
 
--- a/app/Filament/Resources/Playlists/PlaylistResource.php
+++ b/app/Filament/Resources/Playlists/PlaylistResource.php
@@ -207,6 +207,11 @@ class PlaylistResource extends Resource
                     ->sortable()
                     ->poll(fn($record) => $record->status === Status::Processing || $record->status === Status::Pending ? '3s' : null)
                     ->toggleable(),
+                ProgressColumn::make('vod_progress')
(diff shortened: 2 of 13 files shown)


Serph91P:
@sparkison another one, dunno if you like it :smile:

sparkison:
I forgot that we actually import all the **Live** and **VOD** content in one go (as part of the same batching process), so this doesn't really make sense. The series is imported as a separate task so it works there. For this one, I'd need to pull out the processing to make Live and VOD truly separate. As it stands now, they're tightly integrated with the sync process.

Serph91P:
> I forgot that we actually import all the **Live** and **VOD** content in one go (as part of the same batching process), so this doesn't really make sense. The series is imported as a separate task so it works there. For this one, I'd need to pull out the processing to make Live and VOD truly separate. As it stands now, they're tightly integrated with the sync process.

did i miss something I thought I did that?

PR #523, opened 2025-12-09
feat: Add PROXY_URL_OVERRIDE_INCLUDE_LOGOS option

Serph91P:
Add new environment variable to control whether logo URLs should use the PROXY_URL_OVERRIDE setting or the default app URL.

This is useful for Plex users who need HTTPS for logo images. When using a domain with HTTPS for the frontend but PROXY_URL_OVERRIDE points to a local HTTP address, Plex may not load the logos due to HTTPS requirements.

By setting PROXY_URL_OVERRIDE_INCLUDE_LOGOS=false, users can keep the stream proxy override for local access while logos still use the HTTPS domain URL that Plex requires.

Default: true (existing behavior preserved)

### DIFF
--- a/app/Http/Controllers/LogoProxyController.php
+++ b/app/Http/Controllers/LogoProxyController.php
@@ -19,7 +19,7 @@ class LogoProxyController extends Controller
     {
         try {
             // Decode the URL
-            $originalUrl = base64_decode(strtr($encodedUrl, '-_', '+/').str_repeat('=', (4 - strlen($encodedUrl) % 4) % 4));
+            $originalUrl = base64_decode(strtr($encodedUrl, '-_', '+/') . str_repeat('=', (4 - strlen($encodedUrl) % 4) % 4));
 
             // Validate the decoded URL
             if (! filter_var($originalUrl, FILTER_VALIDATE_URL)) {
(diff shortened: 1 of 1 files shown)


Serph91P:
@sparkison another one dunno if you like this but would be great for plex.

sparkison:
Great idea! I didn't think of this, but totally makes sense. I'll get this merged in shortly...

Serph91P:
> Great idea! I didn't think of this, but totally makes sense. I'll get this merged in shortly...

Could we do something similar for the proxy? Because for me the proxy wont work since both urls are http and not https. Only the front-end is.

sparkison:
> Could we do something similar for the proxy? Because for me the proxy wont work since both urls are http and not https. Only the front-end is.

In that case wouldn't it completely disable the `PROXY_URL_OVERRIDE`? I could be missing something though! So let me know, I'm definitely open to adding more options if needed.

Serph91P:
> > Could we do something similar for the proxy? Because for me the proxy wont work since both urls are http and not https. Only the front-end is.
> 
> In that case wouldn't it completely disable the `PROXY_URL_OVERRIDE`? I could be missing something though! So let me know, I'm definitely open to adding more options if needed.

This would only be for the frontend you know like disable proxy override for frontend. Defaults to false but you can set to true.

Serph91P:
> > Could we do something similar for the proxy? Because for me the proxy wont work since both urls are http and not https. Only the front-end is.
> 
> In that case wouldn't it completely disable the `PROXY_URL_OVERRIDE`? I could be missing something though! So let me know, I'm definitely open to adding more options if needed.

ignore what I asked, I just realized there is the dynamic option for the proxy url.

PR #530, opened 2025-12-13
feat: implement chunked VOD processing to prevent timeouts

Serph91P:
- ProcessVodChannels now dispatches a job chain instead of processing all channels in one job
- ProcessVodChannelsChunk processes 100 channels per chunk with 10-minute timeout
- ProcessVodChannelsComplete finalizes the VOD sync and sets status to completed
- Single channel processing still works directly without chunking
- Uses Bus::chain() for sequential processing with error handling
- Updates vod_progress incrementally during processing
- Continues processing on individual channel errors instead of failing the entire job

This fixes timeout issues when processing playlists with 16k+ VOD channels.

### DIFF
--- a/app/Jobs/ProcessVodChannels.php
+++ b/app/Jobs/ProcessVodChannels.php
@@ -3,7 +3,6 @@
 namespace App\Jobs;
 
 use App\Enums\Status;
-use App\Events\SyncCompleted;
 use App\Models\Channel;
 use App\Models\Playlist;
 use App\Services\XtreamService;
@@ -204,9 +203,6 @@ class ProcessVodChannels implements ShouldQueue
                         'vod_processing' => false,
                     ],
                 ]);
-
-                // Fire the sync completed event
-                event(new SyncCompleted($playlist));
             })->dispatch();
     }
 }
--- a/app/Jobs/ProcessVodChannelsComplete.php
+++ b/app/Jobs/ProcessVodChannelsComplete.php
@@ -3,7 +3,6 @@
 namespace App\Jobs;
 
 use App\Enums\Status;
-use App\Events\SyncCompleted;
 use App\Models\Playlist;
 use Filament\Notifications\Notification;
 use Illuminate\Contracts\Queue\ShouldQueue;
@@ -53,8 +52,5 @@ class ProcessVodChannelsComplete implements ShouldQueue
             ->body($message)
             ->broadcast($this->playlist->user)
             ->sendToDatabase($this->playlist->user);
-
-        // Fire the playlist synced event
-        event(new SyncCompleted($this->playlist));
     }
 }


sparkison:
From past experience, this is going to cause issues:
```
$channelIds = $query->pluck('id')->toArray();
$chunks = array_chunk($channelIds, self::CHUNK_SIZE);
$totalChunks = count($chunks);
```

this could load 100k's of channel IDs into an array, and when we've done this before it caused out of memory issues on some machines. I'd recommend changing this to use generators so it doesn't need to store large amounts of data (arrays) in memory.

Serph91P:
> From past experience, this is going to cause issues:
> 
> ```
> $channelIds = $query->pluck('id')->toArray();
> $chunks = array_chunk($channelIds, self::CHUNK_SIZE);
> $totalChunks = count($chunks);
> ```
> 
> this could load 100k's of channel IDs into an array, and when we've done this before it caused out of memory issues on some machines. I'd recommend changing this to use generators so it doesn't need to store large amounts of data (arrays) in memory.

Thank you for the feedback. Changed it.

sparkison:
and should `SyncCompleted` be called here? Since this is a metadata and not provider sync we don't typically call the `SyncCompleted`. Let me know if you had a reason to do so though!

Serph91P:
> and should `SyncCompleted` be called here? Since this is a metadata and not provider sync we don't typically call the `SyncCompleted`. Let me know if you had a reason to do so though!

You're right - SyncCompleted should not be called here. VOD metadata sync is not a full provider sync, so:

Auto-merge shouldn't be triggered (channels already exist)
Post-processes with event='synced' are meant for provider syncs, not metadata updates
I only had it in the error catch block for consistency, but that's also incorrect for a metadata sync. I'll remove it now.

fabien4444:
Hello
i am testing your last dev release.
I find very long the process of strm files generation compared to the master branch. I have already made the metadata fetching. I use only the synch .strm files function -> 6 jobs per minute.
For my 10K series, it will take a day to process.
Is there a link with this feature?

Serph91P:
> Hello i am testing your last dev release. I find very long the process of strm files generation compared to the master branch. I have already made the metadata fetching. I use only the synch .strm files function -> 6 jobs per minute. For my 10K series, it will take a day to process. Is there a link with this feature?

This has nothing to do with the speed. I can sync 16000 in matter of minutes. Do you have the request delay enabled? Because this will limit everything.

fabien4444:
> This has nothing to do with the speed. I can sync 16000 in matter of minutes. Do you have the request delay enabled? Because this will limit everything.

I disabled request delay. 
100% CPU and 2 jobs per minute.
I have structure CATEGORIE / SERIES / SEASONS / year + tmdb.

I select some categories in Series -> Bulk action: sync .strm files

I my pendings jobs, i have many [DatabaseNotification] + [BroadcastNotification] + [SyncSeriesStrmFiles]
In the completed jobs: a [SyncSeriesStrmFiles] lasts 7s/10s average to complete. Sometimes it pause for 3 minutes, strange.

I did same some days ago for Movies and generation was fast.

How can i log something for helping?

Serph91P:
> > This has nothing to do with the speed. I can sync 16000 in matter of minutes. Do you have the request delay enabled? Because this will limit everything.
> 
> I disabled request delay. 100% CPU and 2 jobs per minute. I have structure CATEGORIE / SERIES / SEASONS / year + tmdb.
> 
> I select some categories in Series -> Bulk action: sync .strm files
> 
> I my pendings jobs, i have many [DatabaseNotification] + [BroadcastNotification] + [SyncSeriesStrmFiles] In the completed jobs: a [SyncSeriesStrmFiles] lasts 7s/10s average to complete. Sometimes it pause for 3 minutes, strange.
> 
> I did same some days ago for Movies and generation was fast.
> 
> How can i log something for helping?

How many series you got? How many episodes. I just had the same for me it takes about 8h to sync my series which are 2900 series with 96000 episodes. As you already can see 100% CPU this takes is quit i/o heavy.

fabien4444:
Hello
i have 15K series and 378K episodes :) So it's a normal timing i understand, i will wait. It does not finished this morning.

I have additional question about sync:
- I sync my playlist every day, so i can have new movies/series and removed ones.
- On the VOD & Series processing, does the metadata fetching and STRM files apply on the full content or only the new content?
- The removed contents are removed also in database and STRM files are deleted as well?

Thanks again for your effort.

Serph91P:
> I have additional question about sync:
> 
>     * I sync my playlist every day, so i can have new movies/series and removed ones.
> 
>     * On the VOD & Series processing, does the metadata fetching and STRM files apply on the full content or only the new content?
> 
>     * The removed contents are removed also in database and STRM files are deleted as well?

1. The sync process is incremental and intelligent. It does NOT re-process everything from scratch:

When syncing, the system checks if a .strm file already exists for each item
If the file exists and nothing has changed (same path, same URL), it's skipped
If only the URL changed (e.g., proxy URL updated), only the file content is updated
If the path changed (e.g., category or series name changed), the file is renamed
New content gets new .strm files created
So essentially: only new or changed content is processed.

2. Yes. After each sync, the system runs StrmFileMapping::cleanupOrphaned() which:

Finds all .strm file mappings where the source item (VOD/Episode) no longer exists or is disabled
Deletes the actual .strm file from disk
Deletes the database record for that mapping
Cleans up any empty directories that result from file deletions
So when content is removed from your playlist during the daily sync:

fabien4444:
Thanks for your explanation, it's very clear.

- If i delete a content in the database, does it recover/enable after Playlist syncs?
- If i disable a content and check enable new channel in the playlist settings, does it enable again after sync?

Depending the answer, the better way to keep unwanted VOD/Series in the dabase syncronised with the playlist?

PR #532, opened 2025-12-13
Media Server Integration (Emby/Jellyfin)

Grimothy:
💻 Media Server Integration (Emby/Jellyfin)
🌟 Overview
Adds support for syncing content from Emby and Jellyfin media servers as a sidecar integration, allowing users to import their personal media libraries directly into M3U Editor.

This feature enables the creation of standard VOD Channels and Series structures within the application, utilizing the media server as the content source via direct streaming URLs.

🖼️ Screenshots / Media

image

image

image

image

🛠️ Implementation Details
New Components
Migration: 2025_12_13_000000_create_media_server_integrations_table.php

Model: MediaServerIntegration.php

Service: MediaServerService.php - Handles API communication with Emby/Jellyfin

Job: SyncMediaServer.php - Manages background sync operations

Resource: MediaServerIntegrationResource.php - Admin UI

Enum Addition: PlaylistSourceType::Emby and PlaylistSourceType::Jellyfin

Architecture
Sidecar Philosophy: Creates a dedicated Playlist per integration without modifying existing content tables.

Non-intrusive: All imported content flows through standard models (Channel, Group, Series, Season, Episode, Category).

Batch Tracking: Uses import_batch_no for managing sync operations and preventing duplicates.

Queue Support: Sync jobs run via Laravel Horizon for non-blocking operations.

Data Flow
Media Server (Emby/Jellyfin)
   ↓
MediaServerService (API calls)
   ↓
SyncMediaServer Job
   ↓
Playlist → Groups/Categories → Channels/Series → Episodes
   ↓
Xtream API Output
⚙️ User Experience
Setup Process
Navigate to Integrations → Media Servers.

Click Create and configure: Name, Server Type (Emby/Jellyfin), Host, Port, API Key, SSL toggle, and import/genre handling options.

Click Test Connection to verify server accessibility and credentials.

Save the integration and use the Sync Now action.

Key Features Available
✅ Direct Streaming: Uses direct, non-transcoding streaming URLs for optimal bandwidth.

✅ Full Metadata Sync: Imports plot, cast, director, ratings, runtime, and all associated art (Cover, Backdrop, Logo).

✅ Content Structure: Maps Movies to VOD Channels and TV Series to the standard Category/Series/Season/Episode hierarchy.

✅ Integration Management: View sync statistics, last sync time, and linked playlist directly from the integration resource.

✅ Automatic Output: Synced content automatically appears in standard Playlists, VODs, Series views, and the final Xtream API Output.

🧪 Testing
Tested with Emby server containing 287 movies, 85 series, 3,684 episodes.

Verified VOD metadata display (plot, cast, director, ratings) is accurate.

Verified series/season/episode hierarchy and episode descriptions in the admin UI.

Tested connection validation, and both SSL and non-SSL configurations.

🚨 Breaking Changes
None - this is an additive feature.

📝 Configuration
All primary configuration is managed per-integration via the admin UI. No new global configuration is strictly required, but the structure allows for future enhancements via

### DIFF
--- a/app/Http/Controllers/MediaServerProxyController.php
+++ b/app/Http/Controllers/MediaServerProxyController.php
@@ -1,0 +1,282 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Models\MediaServerIntegration;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+/**
+ * MediaServerProxyController - Secure proxy for Emby/Jellyfin content
+ *
+ * This controller proxies requests to media servers, hiding the API key
+ * from external clients (IPTV players). Similar to SchedulesDirectImageProxyController.
+ */
+class MediaServerProxyController extends Controller
+{
+    /**
+     * Proxy an image from the media server.
+     *
+     * Route: /media-server/{integrationId}/image/{itemId}/{imageType?}
+     *
+     * @param Request $request
+     * @param int $integrationId The integration ID
+     * @param string $itemId The media server's item ID
+     * @param string $imageType The image type (Primary, Backdrop, Logo, etc.)
+     * @return Response|StreamedResponse
+     */
+    public function proxyImage(Request $request, int $integrationId, string $itemId, string $imageType = 'Primary')
+    {
+        try {
+            $integration = MediaServerIntegration::find($integrationId);
+
+            if (!$integration) {
+                return response()->json(['error' => 'Integration not found'], 404);
+            }
+
+            if (!$integration->enabled) {
+                return response()->json(['error' => 'Integration is disabled'], 403);
+            }
+
+            // Build cache key for this image
+            $cacheKey = "media_server_image_{$integrationId}_{$itemId}_{$imageType}";
+
+            // Check cache first (cache for 24 hours)
+            $cachedResponse = Cache::get($cacheKey);
+            if ($cachedResponse) {
+                return response($cachedResponse['body'], 200, $cachedResponse['headers']);
+            }
+
+            // Build the media server image URL with authentication
+            $imageUrl = "{$integration->base_url}/Items/{$itemId}/Images/{$imageType}";
+
+            // Fetch the image with authentication
+            $response = Http::withHeaders([
+                'X-Emby-Token' => $integration->api_key,
+                'Accept' => 'image/*',
+            ])->timeout(30)->get($imageUrl);
+
+            if ($response->successful()) {
+                $body = $response->body();
+                $contentType = $response->header('Content-Type', 'image/jpeg');
+
+                // Prepare headers for the proxied response
+                $headers = [
+                    'Content-Type' => $contentType,
+                    'Content-Length' => strlen($body),
+                    'Cache-Control' => 'public, max-age=86400', // Cache for 24 hours
+                    'X-Proxied-From' => 'MediaServer',
+                ];
+
+                // Cache the successful response for 24 hours
+                Cache::put($cacheKey, [
+                    'body' => $body,
+                    'headers' => $headers,
+                ], now()->addHours(24));
+
+                Log::debug('Successfully proxied media server image', [
+                    'integration_id' => $integrationId,
+                    'item_id' => $itemId,
+                    'image_type' => $imageType,
+                    'content_type' => $contentType,
+                    'size_bytes' => strlen($body),
+                ]);
+
+                return response($body, 200, $headers);
+            }
+
+            Log::warning('Failed to fetch media server image', [
+                'integration_id' => $integrationId,
+                'item_id' => $itemId,
+                'image_type' => $imageType,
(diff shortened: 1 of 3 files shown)


sparkison:
@Grimothy, any concerns with the `?api_key=` in the generated URLs? Just want to make sure, since this will expose the API key to the Xtream API and other endpoints and outputs.

Wasn't sure if we should use an internal proxy route to fetch and return the results instead? We do something similar for the SchedulesDirect integration. We cache the images in that case, which I don't think we need here. Thinking adding a media server media proxy controller should be able to fetch the media content from the server and return it to the client.

Let me know what you think! And thanks for all the great work! This is looking pretty solid 💪

fabien4444:
Hello
Is this feature working in a release i could test?

Grimothy:
> @Grimothy, any concerns with the `?api_key=` in the generated URLs? Just want to make sure, since this will expose the API key to the Xtream API and other endpoints and outputs.
> 
> Wasn't sure if we should use an internal proxy route to fetch and return the results instead? We do something similar for the SchedulesDirect integration. We cache the images in that case, which I don't think we need here. Thinking adding a media server media proxy controller should be able to fetch the media content from the server and return it to the client.
> 
> Let me know what you think! And thanks for all the great work! This is looking pretty solid 💪

Oh wow rookie move. I will look at it :P

Grimothy:
> Hello Is this feature working in a release i could test?
IF you REALLY wanted to test, you would need to pull the code from my github. (Its not merged yet)

Grimothy:
@sparkison the url in the player reports:

image

API Key is also no longer exposed through the web interface:

image

sparkison:
Nicely done! 🎉

Grimothy:
> Nicely done! 🎉

Thank you kind sir. Lets get this tested so I can work on some of the other nonsense  :D

fabien4444:
Hello
Thanks for you feature,  very promising. 
I tested with my playlist to sync with my Emby server but no series no movies is imported. 
The connexion test is ok. 
Must i have a prepared library on Emby before sync processing ?

Grimothy:
> Hello Thanks for you feature, very promising. I tested with my playlist to sync with my Emby server but no series no movies is imported. The connexion test is ok. Must i have a prepared library on Emby before sync processing ?

Yes, you need to have contents in your library. I am looking at a separate bug to this that is not tied to your issue. But yes, you definitely need content in there to sync :D

Grimothy:
Actually not a bug. Just a note: Make sure you set the app url correctly :D

fabien4444:
I have a one library 'music video' for local concerts files. 
Must i have a library for Series and another one for Movies ?

fabien4444:
Hello
I have created 2 libraries, one for Series, one for Movies. They are empty. I want to full them with my provider contents thanks to your feature. 
Unfortunatly, the sync processed 0 content.
I misses something?

Pictures of Emby libraries and MP3_editor snapshots:
library
editor

Grimothy:
> Hello I have created 2 libraries, one for Series, one for Movies. They are empty. I want to full them with my provider contents thanks to your feature. Unfortunatly, the sync processed 0 content. I misses something?
> 
> Pictures of Emby libraries and MP3_editor snapshots: library editor

OOOOOHHHHHHHHHHHHHHHH. No its the opposite way I am afraid. I think what you may be looking for is the .strm feature that this application has. I think Emby has a new feature if you pay for premier  that you can import .strm files. 


The workflow for this is:


Emby --> M3ue ---->Client

fabien4444:
Oups my mistake, sorry. It's only one way Emby -> m3u editor :(

Is it possible to develop the other way? It would be very useful and allow to avoid the STRM files.

Grimothy:
> Oups my mistake, sorry. It's only one way Emby -> m3u editor :(
> 
> Is it possible to develop the other way? It would be very useful and allow to avoid the STRM files.

I dont think we need to develop anything. There are a few options you can do with both. I am not super familiar with the web streams feature 

image


"A new plugin is in the Emby plugin catalog for Emby Premiere users, largely developed by @softworkz. This plugin requires Emby Server version 4.9.1.27+, so unless you're running that, you will not see it.

WebStreams

This is designed to be a better version of the IPTV plugin. You can now import both single streams as well as M3U lists into Emby Server. Eventually we may also add RSS support and merge the podcasts plugin into this. Beyond RSS, there is also the potential to support other formats if there's enough demand for them.

Features

Import both single streams and M3U lists into Emby Server
TV episodes within M3U are displayed with Series and Season hierarchies, just like regular libraries (make sure to select the appropriate content type when adding a source)
Rich metadata support using MovieDb and TVDB (make sure to select the appropriate content type when adding a source)
Ability to filter what gets imported
Known Issues

We need to add more visibility into the server's background refresh queue. You can see the progress of it on the webstreams config page. Currently if you restart the server in the middle of this, then you will need to run the refresh channels scheduled task."
https://emby.media/community/index.php?/topic/142334-new-webstreams-plugin/

Theoretically you can use the VOD stream file settings and the Series stream file settings under Settings > Sync options.
This will give you options to generate .strm files which if you didnt have this nice feature would be a pain in the butt.  The .strm files for series and vods are placed in a directory that you choose on the m3ue container.  You could create a mount point for the container that is accessible from both your emby server and your m3ue. You would then create a playlist and point it to where the .strm files are. 




You content will appear :D

PR #538, opened 2025-12-15
fix: Add fallback to original groups/categories in Custom Playlist Xtream API

Serph91P:
When channels/series are added to a Custom Playlist without assigning a custom group/category tag, the Xtream API categories endpoints now correctly fall back to the original group_id/category_id.

Previously, only custom tags were returned in get_live_categories, get_vod_categories, and get_series_categories. This caused channels without custom tags to appear in streams with a category_id that wasn't listed in the categories response.

Changes:
- get_live_categories: Include original groups for channels without tags
- get_vod_categories: Include original groups for VOD without tags
- get_series_categories: Include original categories for series without tags

All three endpoints now check for channels/series that don't have custom tags assigned and include their original group/category as a fallback, avoiding duplicate category_ids.

### DIFF
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -915,6 +915,7 @@ class XtreamApiController extends Controller
                     ->where('is_vod', false)
                     ->pluck('id');
 
+                // Get custom tags assigned to channels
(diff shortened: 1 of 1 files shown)


sparkison:
Looks good!

PR #539, opened 2025-12-15
feat: Add name filtering option for STRM file sync

Serph91P:
Add ability to remove specific words/symbols from folder and file names when generating .strm files for Series and VOD content.

Features:
- New 'Name Filtering' section in Preferences > Sync Options
- Configurable patterns to remove (e.g. 'DE • ', 'EN |', '[HD]')
- Applies to category names, series names, episode titles (Series)
- Applies to group names and VOD titles (VOD)
- Disabled by default, can be enabled per content type

Use case: Remove country prefixes like 'DE • Action' -> 'Action'

Files changed:
- app/Settings/GeneralSettings.php: New settings properties
- app/Filament/Pages/Preferences.php: UI for name filtering
- app/Jobs/SyncSeriesStrmFiles.php: Apply filtering logic
- app/Jobs/SyncVodStrmFiles.php: Apply filtering logic
- database/settings/2025_12_15_*: Migration for new settings

### DIFF
--- a/app/Filament/Pages/Preferences.php
+++ b/app/Filament/Pages/Preferences.php
@@ -673,6 +673,22 @@ class Preferences extends SettingsPage
                                                     ]),
                                             ])
                                             ->hidden(fn($get) => ! $get('stream_file_sync_enabled')),
+                                        Fieldset::make('Name Filtering')
+                                            ->columnSpanFull()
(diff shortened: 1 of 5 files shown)


sparkison:
Another great addition, thanks for the assist! 🙏

PR #541, opened 2025-12-15
Update PlaylistAliasResource.php

Aldione:
Add support for multiple Xtream API configurations in Custom Playlists.

This update introduces a dynamic selector (1–6) for the number of playlists when using a Custom Playlist, and renders independent Xtream API configuration tabs for each playlist.

The implementation preserves backward compatibility with existing single-config setups by normalizing legacy Xtream configurations to the new array-based format.

It also improves robustness when using the stream proxy by safely handling incomplete configurations and ensuring the correct Xtream credentials are used based on the source playlist URL.

### DIFF
--- a/app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php
+++ b/app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php
@@ -35,6 +35,8 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Validation\Rule;
 use App\Traits\HasUserFiltering;
+use Filament\Schemas\Components\Tabs;
+use Filament\Schemas\Components\Tabs\Tab;
 
 class PlaylistAliasResource extends Resource
 {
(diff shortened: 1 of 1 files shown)


sparkison:
I don't quite understand the purpose of this PR? I've been looking at it, and can't quite wrap my head around its purpose? Could you explain a bit more why we should add this?

Aldione:
This PR extends Custom Playlist Aliases to support multiple source playlists (1–6),
each with its own Xtream API configuration.

Why this is needed

Currently, a Custom Playlist Alias can only be linked to a single Xtream API
configuration. In real-world IPTV setups this is a limitation because users often
merge multiple providers into a single logical Custom Playlist (for example: one
provider for live TV, another for VOD, or a backup provider).

Each source playlist usually requires different credentials and endpoints.
Without this change, users are forced to create multiple aliases or duplicate
playlists, increasing maintenance complexity and the risk of configuration errors.

What this PR changes

- When a Playlist Alias points to a Custom Playlist, the user can select how many
  source playlists it contains (1–6).
- The UI dynamically generates separate Xtream API configuration tabs for each
  source playlist.
- Existing aliases using a single Xtream API configuration remain fully functional
  (backward compatible).
- When Stream Proxy is enabled, the system can automatically associate the correct
  credentials by matching the playlist URL with the corresponding Xtream API
  configuration.

Why this logic lives in PlaylistAliasResource

The alias is already responsible for authentication, proxying, and source
resolution. Custom Playlists act as an aggregation layer, while the alias defines
how each source is accessed. Keeping this logic here avoids duplicating Xtream
configuration across multiple aliases or playlists.

Scope & safety

- Fully backward compatible.
- Hard limit of 6 playlists to keep complexity bounded.
- No behavior change for non-Custom Playlists.

In short, this PR removes a real limitation for advanced users while keeping the
default/simple use case exactly the same.

PR #542, opened 2025-12-15
Update PlaylistAlias.php

Aldione:
Refactor Xtream configuration handling to support multiple provider setups.

This change introduces helper methods to normalize legacy and multi-config Xtream API definitions, allowing a Playlist Alias to work with one or more Xtream providers transparently.

URL transformation logic has been updated to safely select the appropriate Xtream configuration based on the source playlist URL, while preserving existing behavior for single-config aliases and avoiding credential conflicts when multiple configurations are present.

### DIFF
--- a/app/Models/PlaylistAlias.php
+++ b/app/Models/PlaylistAlias.php
@@ -29,6 +29,53 @@ class PlaylistAlias extends Model
         'strict_live_ts' => 'boolean',
     ];
 
+    public function getXtreamConfigs(): array
+    {
+        $raw = $this->xtream_config;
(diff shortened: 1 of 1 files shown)


Aldione:
Is there any issue?

sparkison:
While I might revisit at a later time, this isn't something I'm interested in including atm. Thank you for the request!

sparkison:
Ok! Scratch my last comment, I now understand this functionality and will get it pulled into the core app! 🎉 
thanks for the work @Aldione 🙏

PR #549, opened 2025-12-16
feat(strm): improve STRM file naming with Trash Guides format support

Serph91P:
- Fix VOD name filtering bug: use group->name instead of group model object
- Add TMDB ID to VOD title folder for Trash Guides compatibility
- Add TVDB/TMDB/IMDB ID to Series folder names (priority: TVDB > TMDB > IMDB)
- Smart TMDB ID placement: folder name when title folder enabled, filename otherwise
- Update Preferences preview to reflect actual naming logic
- Fix group name preview to use group->name instead of model object


sparkison:
Looks good to me! 😉

PR #552, opened 2025-12-17
feat(strm): implement intelligent STRM sync with file tracking

Serph91P:
- Add StrmFileMapping model to track .strm file locations in database
- Implement intelligent rename instead of delete/create when path changes
- Automatic URL updates when proxy settings or base URL changes
- Clean up empty directories after file moves
- Add Trash Guides format support for VOD and Series folders:
  - VOD: {Movie Title} ({Year}) {tmdb-ID}
  - Series: {Series Name} ({Year}) {tvdb-ID} (falls back to tmdb/imdb)
- Smart TMDB ID placement: folder name when title folder enabled, filename otherwise
- Fix name filtering bug: use group->name instead of group model object
- Update Preferences preview to reflect actual naming logic

Migration: Creates strm_file_mappings table for tracking sync state

### DIFF
--- a/app/Jobs/SyncSeriesStrmFiles.php
+++ b/app/Jobs/SyncSeriesStrmFiles.php
@@ -3,8 +3,8 @@
 namespace App\Jobs;
 
 use App\Facades\ProxyFacade;
+use App\Models\Episode;
 use App\Models\Series;
-use App\Models\SeriesEpisode;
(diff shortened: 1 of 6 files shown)


sparkison:
Love it! This is great, thanks for the effort as always, much appreciated! 🙏

PR #555, opened 2025-12-17
update PlaylistAliases.php

Aldione:
Fix PlaylistAlias counters for multiple aliases from same Custom Playlist

Explicitly set the local parent key (custom_playlist_id) in PlaylistAlias
belongsToMany relationships to ensure pivot joins use the correct
Custom Playlist reference.

This fixes Live/VOD/Series counters showing 0 on second and subsequent
aliases created from the same Custom Playlist.


sparkison:
Since this was targeted at `master`, I've went ahead and manually copied over the update to `dev` 😉

PR #560, opened 2025-12-18
Fix: prevent original playlist categories leaking into CustomPlaylist…

Aldione:
… Xtream responses

Root cause:
CustomPlaylist Xtream endpoints were falling back to original group_id / category_id when channels or series had no custom tag assigned. This caused original playlist categories (often from the second merged playlist) to appear in Xtream clients, even when custom groups were explicitly defined.

What changed:
- Removed fallback logic that merged original groups/categories into get_live_categories, get_vod_categories and get_series_categories for CustomPlaylist.
- Updated get_live_streams and get_vod_streams to filter strictly by custom tag IDs (no fallback to group_id).
- Updated get_series filtering to rely exclusively on custom category tags.
- Ensured category_id assignment for CustomPlaylist uses only custom tag IDs; untagged items now return category_id = "0".
- Added temporary debug logs to validate tag usage and detect untagged items.

Result:
Xtream clients now receive a consistent and deterministic category list for CustomPlaylist. Only user-defined custom groups are exposed, original playlist groups are fully ignored, and category IDs remain stable and Xtream-compatible.

### DIFF
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -475,22 +475,18 @@ class XtreamApiController extends Controller
             // Apply category filtering if category_id is provided
             if ($categoryId && $categoryId !== 'all') {
                 if ($playlist instanceof CustomPlaylist || ($playlist instanceof PlaylistAlias && $playlist->custom_playlist_id)) {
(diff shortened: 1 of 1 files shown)


sparkison:
I've implemented this in the repo! Closing as it conflicts with my updated implementation.

PR #561, opened 2025-12-18
Add selectable active Xtream config to PlaylistAlias

Aldione:
Introduce support for selecting which Xtream configuration is treated as the
primary provider for a Playlist Alias.

Changes:
- Added selected_xtream_config_index to persist the chosen provider.
- Updated primary Xtream config resolution to use the selected index
  instead of always defaulting to the first configuration.
- Ensured safe fallback to index 0 when the selected provider is missing
  or out of range.
- Preserved backward compatibility by keeping index 0 as the default.

Result:
Playlist Aliases can now correctly use different providers with independent
expiry dates and status information.


sparkison:
I've implemented this in the repo! Closing as this is no longer relevant 😉

PR #562, opened 2025-12-18
Improve Playlist Alias provider selection UI and stability

Aldione:
Enhance the Playlist Alias edit form to properly support multiple Xtream
providers in a clear and stable way.

Changes:
- Added an explicit selector to choose the active Xtream provider using
  human-readable playlist names.
- Fixed issues where provider options could collapse after saving or editing.
- Normalized xtream_config state when loading existing aliases to prevent
  credential loss.
- Ensured the active provider selection resets safely to the default
  when providers are added or removed.

Result:
Editing Playlist Aliases no longer requires re-selecting Custom Playlists
or re-entering provider credentials, and the active provider selection
remains consistent across edits.


sparkison:
I've just added a priority selector to the credentials to keep it simple 😉

PR #563, opened 2025-12-18
Add selected_xtream_config_index to playlist_aliases

Aldione:
I forgot to add this for the expiry date


sparkison:
ah! I was wondering 😄

I came up with another solution that's a little simpler (setting the sort priority). End results should be the same 😉

PR #565, opened 2025-12-18
Change label from 'Xtream API Password' to 'Xtream API Username'

Aldione:


Aldione:
I saw you already change it

PR #571, opened 2025-12-20
feat: Add Background Jobs transparency feature

Serph91P:
- Add new Jobs page (Tools > Background Jobs) with progress tracking table
- Add JobProgress model and migration for tracking job state
- Add TracksJobProgress trait for easy job integration
- Add job indicator in header showing active job count with animation
- Add queue warnings on sync action modals when jobs are already running

Jobs now tracked:
- Playlist sync (ProcessM3uImport)
- EPG sync (ProcessEpgImport)
- EPG mapping (MapPlaylistChannelsToEpg)
- Series metadata sync (ProcessM3uImportSeries)
- Series STRM sync (SyncSeriesStrmFiles)
- VOD metadata sync (ProcessVodChannels)
- VOD STRM sync (SyncVodStrmFiles)

Series Processing improvements:
- Split into 3 separate toggles (matching VOD): Fetch metadata, Sync stream files, Include in M3U
- Add auto_sync_series_stream_files field to playlists
- Auto-trigger Series STRM Sync after Series import when enabled

Progress update optimizations:
- VOD STRM sync: Update every 100 items + final update before completion
- Series STRM sync: Update every 10 series + lazy relation loading

### DIFF
--- a/app/Filament/Pages/Jobs.php
+++ b/app/Filament/Pages/Jobs.php
@@ -1,0 +1,166 @@
+<?php
+
+namespace App\Filament\Pages;
+
+use App\Models\JobProgress;
+use Filament\Actions\Action;
+use Filament\Notifications\Notification;
+use Filament\Pages\Page;
+use Filament\Support\Icons\Heroicon;
+use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Concerns\InteractsWithTable;
+use Filament\Tables\Contracts\HasTable;
+use Filament\Tables\Table;
+use Illuminate\Contracts\Support\Htmlable;
+use RyanChandler\FilamentProgressColumn\ProgressColumn;
+
+class Jobs extends Page implements HasTable
+{
+    use InteractsWithTable;
+
+    protected static string|\BackedEnum|null $navigationIcon = 'heroicon-o-queue-list';
+
+    protected string $view = 'filament.pages.jobs';
+
+    protected static ?string $navigationLabel = 'Background Jobs';
+
+    protected static ?string $title = 'Background Jobs';
+
+    protected ?string $subheading = 'Monitor the progress of background jobs like playlist syncs, EPG imports, and metadata fetching.';
+
+    public static function getNavigationSort(): ?int
+    {
+        return 1;
+    }
+
+    public static function getNavigationGroup(): ?string
+    {
+        return 'Tools';
+    }
+
+    public static function getNavigationBadge(): ?string
+    {
+        $count = JobProgress::active()->count();
+
+        return $count > 0 ? (string) $count : null;
+    }
+
+    public static function getNavigationBadgeColor(): ?string
+    {
+        return 'primary';
+    }
+
+    public function getTitle(): string|Htmlable
+    {
+        return 'Background Jobs';
+    }
+
+    protected function getActions(): array
+    {
+        return [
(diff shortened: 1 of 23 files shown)


sparkison:
If you want to submit a separate PR for the Series sync separation (the addition of the `auto_sync_series_stream_files` for Playlists I'll get that merged.

The original thinking is that you wouldn't sync without also fetching metadata, otherwise the episodes would be empty. With VOD, it doesn't matter as much since we don't need to fetch additional data to get episodes for the different URLs.

Still open to splitting up Series like VOD if you feel it makes sense though 🤷 

For the job tracking, I'd rather utilize a plugin, something like this (which basically does the same thing): https://github.com/mooxphp/jobs/tree/v4.0.0-beta.1

For the reason above, I'm gonna close this one...

Serph91P:
> If you want to submit a separate PR for the Series sync separation (the addition of the `auto_sync_series_stream_files` for Playlists I'll get that merged.
> 
> The original thinking is that you wouldn't sync without also fetching metadata, otherwise the episodes would be empty. With VOD, it doesn't matter as much since we don't need to fetch additional data to get episodes for the different URLs.
> 
> Still open to splitting up Series like VOD if you feel it makes sense though 🤷
> 
> For the job tracking, I'd rather utilize a plugin, something like this (which basically does the same thing): https://github.com/mooxphp/jobs/tree/v4.0.0-beta.1
> 
> For the reason above, I'm gonna close this one...

Hi @sparkison,

I've tested the `moox/jobs` plugin (v4.0.0-beta.3) as you suggested, but unfortunately I encountered a critical error that prevents the app from loading:

```
Error: Object of class Moox\Jobs\JobsPlugin could not be converted to string
Location: vendor/laravel/framework/src/Illuminate/Support/Str.php:528
```

## What I tried

- Installed via `composer require moox/jobs:^4.0.0-beta.1`
- Ran `php artisan mooxjobs:install` (which partially worked but broke the app)
- Manual installation following the docs (migrations, assets, plugin registration)
- All requirements are met: Laravel 12.42, Filament 4.3.1, PHP 8.4.14

## The error occurs

When registering the plugins in `AdminPanelProvider.php`:

```php
->plugins([
    \Moox\Jobs\JobsPlugin::make(),
    \Moox\Jobs\JobsWaitingPlugin::make(),
    \Moox\Jobs\JobsFailedPlugin::make(),
    \Moox\Jobs\JobsBatchesPlugin::make(),
])
```

The app returns a 500 error immediately after adding these plugin registrations.

## Current Setup

- Laravel Horizon is already installed and running perfectly
- Queue driver: Redis
- Horizon provides comprehensive job monitoring at `/horizon`

## Proposed Solutions

Since moox/jobs v4 is still in beta and appears to have compatibility issues, I'd like to suggest some alternatives:

1. **Use Laravel Horizon as primary job monitoring** (already working)
   - Add a navigation link to Horizon in Filament's Tools menu
   - This is the official Laravel tool for Redis-based queues
   
2. **Wait for stable release** of moox/jobs v4 with proper Filament 4 support
   
3. **Keep the custom implementation** from the `feature/background-jobs-transparency` branch
   - Already integrated into Filament
   - Works with the existing Horizon setup

I'm happy to implement option 1 (Horizon link in Filament navigation) as a pragmatic solution until moox/jobs v4 reaches a stable release. Would that work for you?

Let me know your preference.

PR #588, opened 2025-12-30
Add Channel Management Features

Italianar15:
# Fix: Synchronize M3U and EPG Generation & Add Channel Management Features

## Summary
This PR addresses critical issues regarding Emby/Jellyfin compatibility (duplicate channel names, M3U/EPG synchronization) and introduces new channel management features for the Admin Panel.

The key changes enforce strict parity between `PlaylistGenerateController` and `EpgGenerateController`, and add powerful bulk actions for managing channel ordering.

### 1. Synchronized Sort Order
- **Issue**: The M3U generator sorts channels by `Group -> Sort Order -> Channel Number -> Title`, but the EPG generator was missing the Group sort. This caused channels to be processed in a different order, leading to ID mismatches.
- **Fix**: Updated `EpgGenerateController` query to `leftJoin` the `groups` table and apply the exact same `orderBy` priority (`groups.sort_order` first).

### 2. Robust Deduplication Strategy
- **Issue**: Emby treats identical channel names as the same channel. Without unique `tvg-id`s, it merges or hides duplicates.
- **Fix**: Implemented identical deduplication logic in both controllers:
    - **Names**: Appends `(2)`, `(3)`, etc. to duplicate channel names.
    - **IDs**: Appends `.2`, `.3`, etc. to duplicate `tvg-id`s.
- **Improvement**: Moved the deduplication check to **after** the Regex character stripping. This prevents collisions where distinct IDs (e.g., `Ch-1` and `Ch_1`) would become identical (`Ch1`) after stripping.

### 3. Missing ID Fallback
- **Fix**: If a channel has no `tvg-id`, a robust fallback is generated using the filename from the stream URL or an MD5 hash of the URL. This ensures every channel has a unique identifier required by EPG players.

### 1. Channel Recounting (Group Resource)
- **Single Group Recount**: Added a `Recount Channels` action to the Group table. Allows users to specify a start number and sequentially renumbers all channels in that group based on their current sort order.
- **Bulk Group Recount**: Added a bulk action to recount channels across multiple selected groups. The feature respects the group sort order, ensuring channels are numbered continuously from the first group to the last.

### 2. Alpha Sorting
- **Group Alpha Sort**: Added a `Sort Alpha` action to the Group table. This automatically reorders all channels within the group alphabetically by title (respecting custom titles) and updates their `sort` index.

## Affected Files
- `app/Http/Controllers/PlaylistGenerateController.php`
- `app/Http/Controllers/EpgGenerateController.php`
- `app/Filament/Resources/Groups/GroupResource.php`

## Verification
- **Emby Sync**: Tested by importing the generated M3U and XMLTV files into Emby. Confirmed that all channels (including duplicates) are preserved and distinctly named, and EPG data maps correctly.
- **Channel Management**: Verified the Recount and Sort Alpha actions in the Admin Panel function as expected, correctly re-sequencing channels in the database.

### DIFF
--- a/app/Filament/Resources/Playlists/PlaylistResource.php
+++ b/app/Filament/Resources/Playlists/PlaylistResource.php
@@ -343,7 +343,7 @@ class PlaylistResource extends Resource
                                 ->send();
                         })
                         ->disabled(fn($record): bool => $record->isProcessing())
-                        ->hidden(fn($record): bool => !$record->xtream)
+                        ->hidden(fn($record): bool => ! $record->xtream)
                         ->requiresConfirmation()
                         ->icon('heroicon-o-arrow-down-tray')
                         ->modalIcon('heroicon-o-arrow-down-tray')
@@ -368,7 +368,7 @@ class PlaylistResource extends Resource
                                 ->send();
                         })
                         ->disabled(fn($record): bool => $record->isProcessing())
-                        ->hidden(fn($record): bool => !$record->xtream)
+                        ->hidden(fn($record): bool => ! $record->xtream)
                         ->requiresConfirmation()
                         ->icon('heroicon-o-arrow-down-tray')
                         ->modalIcon('heroicon-o-arrow-down-tray')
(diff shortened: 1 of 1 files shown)


sparkison:
Thanks for the contribution! It has been merged into `dev` 😉

PR #596, opened 2025-12-31
perf: Optimize series sync to prevent resource exhaustion

Serph91P:
Summary

Implemented configurable sync performance profiles with automatic hardware detection to prevent system resource exhaustion during large playlist syncs.

Problem

When syncing playlists with thousands of series/episodes, the system would dispatch thousands of individual jobs (e.g., 2900+ SyncSeriesStrmFiles jobs), causing CPU spikes to 100% and system freezes.

Solution

Bulk Processing Optimization

Changed from dispatching one job per series to processing all series in a single bulk job
Series metadata sync now dispatches ONE SyncSeriesStrmFiles job at the end instead of per-series
Added memory management with unsetRelations() and gc_collect_cycles() between chunks
ResourceManager Service

New service that auto-detects available system RAM and CPU cores
Reads from meminfo and handles Docker cgroups limits (v1 and v2)
Provides appropriate chunk sizes based on detected hardware
Performance Profiles

Low (1-2 GB RAM): metadata=5, strm=10, cleanup=250
Medium (2-4 GB RAM): metadata=15, strm=35, cleanup=500
High (4-8 GB RAM): metadata=30, strm=75, cleanup=1000
Ultra (8+ GB RAM): metadata=50, strm=100, cleanup=2000
Auto mode detects optimal profile based on available resources
User Configuration

Added settings UI in Preferences > Sync Options
Users can select a fixed profile or use auto-detection
Custom chunk sizes can be configured for fine-tuning
Applied to All Sync Jobs

ProcessM3uImportSeriesEpisodes (series metadata)
SyncSeriesStrmFiles (series STRM files)
ProcessVodChannels (VOD metadata)
SyncVodStrmFiles (VOD STRM files)
Results

CPU usage reduced from 100% (system freeze) to ~10-25%
RAM usage stable at ~5-10%
No more system crashes during large syncs
Sync still completes in reasonable time with configurable performance

### DIFF
--- a/app/Jobs/ProcessVodChannels.php
+++ b/app/Jobs/ProcessVodChannels.php
@@ -5,6 +5,7 @@ namespace App\Jobs;
 use App\Enums\Status;
 use App\Models\Channel;
 use App\Models\Playlist;
+use App\Services\ResourceManager;
 use App\Services\XtreamService;
 use App\Traits\ProviderRequestDelay;
 use Filament\Notifications\Notification;
@@ -43,7 +44,7 @@ class ProcessVodChannels implements ShouldQueue
     /**
      * Execute the job.
      */
-    public function handle(XtreamService $xtream): void
+    public function handle(XtreamService $xtream, ResourceManager $resourceManager): void
     {
         $playlist = $this->playlist;
         if ($playlist === null) {
@@ -60,8 +61,8 @@ class ProcessVodChannels implements ShouldQueue
             return;
         }
 
-        // For bulk processing, use chunked approach
-        $this->processVodChannelsInChunks($playlist);
+        // For bulk processing, use chunked approach with ResourceManager
+        $this->processVodChannelsInChunks($playlist, $resourceManager);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sparkison:
hmm, I'm thinking this is started to get more complex than it needs to be. This syncing was working fine before this optimized implementation and used low resources.

we need to find a way to better optimize it without adding a bunch of variables to try and squeeze as much as we can out of the system it's running on - it shouldn't be this process intense to begin with.

I'll do some investigating on my side as well to see what I can find, but going to close this one down...

Serph91P:
> hmm, I'm thinking this is started to get more complex than it needs to be. This syncing was working fine before this optimized implementation and used low resources.
> 
> we need to find a way to better optimize it without adding a bunch of variables to try and squeeze as much as we can out of the system it's running on - it shouldn't be this process intense to begin with.
> 
> I'll do some investigating on my side as well to see what I can find, but going to close this one down...

That's too bad. But thanks for the feedback.

PR #597, opened 2025-12-31
fix: Dispatch single bulk STRM sync job instead of per-series

Serph91P:
Previously, when syncing series metadata, a separate SyncSeriesStrmFiles job was dispatched for each series (2900+ jobs for large playlists). This overwhelmed the queue and caused high CPU/memory usage.

Now in bulk mode:
- Per-series sync dispatch is disabled during metadata fetch
- ONE SyncSeriesStrmFiles job is dispatched at the end
- The bulk job processes all series internally with chunking

This dramatically reduces queue pressure and resource usage.


Serph91P:
@sparkison then maybe this is more like what you imagined?

PR #598, opened 2025-12-31
Fix: preserve alias credentials on failover

Aldione:
When a channel is served under a PlaylistAlias, primary URLs are correctly transformed via PlaylistUrlService.

Failover URLs were incorrectly generated using each failover channel’s getEffectivePlaylist(), causing the alias context to be lost and reverting to original playlist credentials.

Updated failover URL generation to reuse the original playlist context (including aliases).

Updated resolver-based failover to resolve and apply the original context via playlist_uuid.

Added regression coverage by ensuring non-alias playlists keep the same behavior.

### DIFF
--- a/app/Services/M3uProxyService.php
+++ b/app/Services/M3uProxyService.php
@@ -3,6 +3,7 @@
 namespace App\Services;
 
 use App\Facades\ProxyFacade;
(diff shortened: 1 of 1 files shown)


sparkison:
Ah! Good catch! Merging now...

PR #626, opened 2026-01-10
Add per-user expiration with date & time for PlaylistAuth and PlaylistAlias

Aldione:
This change introduces per-account expiration support with full date and time precision.

- Added `expires_at` (datetime) to PlaylistAuth and PlaylistAlias.
- Authentication now denies access when the expiration date/time is reached.
- Expiration is enforced consistently across:
  - Xtream API endpoints (player_api.php, get.php, etc.)
  - Stream endpoints (live, movie, series, timeshift)
- Supports hour/minute-level expiration (e.g. 22:34).
- Maintains backward compatibility with legacy owner_auth login.
  - Legacy access can be optionally controlled via a PlaylistAuth override.
- Admin UI updated to allow setting expiration date/time.
- Timezone-aware using application timezone configuration.

This enables precise, controlled access management without interrupting active sessions.


Aldione (commented):
editted

PR #631, opened 2026-01-11
Fix: Preserve Xtream username in m3u-proxy client sessions

Aldione:
### 🐛 Problem
Active streams were correctly detected by m3u-proxy, but when opening “Show Clients” in the stream monitor, the Username column always displayed `-`.

Root cause:
- IPTV clients do not forward custom headers (e.g. X-Username) when following HTTP redirects (302/301).
- The Laravel backend redirected clients to the Python m3u-proxy without preserving the authenticated Xtream username.
- The proxy therefore registered client sessions with username = null.

---

### ✅ Solution
This PR fixes end-to-end client traceability by:

- Appending trace parameters to m3u-proxy URLs in Laravel:
  - username: authenticated Xtream username
  - client_id: unique per-connection UUID to prevent CGNAT / IP collisions
- Extending the Python m3u-proxy client detection to:
  - Read username from the query string as a fallback when headers are missing
  - Correctly associate usernames with each individual client session

---

### 🎯 Result
- The Username column now shows the real user(s) watching each stream
- Multiple users can watch the same stream simultaneously, each tracked independently
- Robust against:
  - shared IPs (CGNAT)
  - identical User-Agents
  - reconnections
- Minimal changes, no breaking behavior

---

### 🔍 Example

Before:
Stream: Channel 101  
Clients:  
- Username: -

After:
Stream: Channel 101  
Clients:  
- Username: john_doe  
- Username: alice_tv  

---

### 🧩 Notes
- Backward compatible
- No database or schema changes
- Works for live, VOD, series and HLS streams


sparkison:
Was running into some merge issues and other problems so I'm gonna close this one and implement directly - added here: https://github.com/sparkison/m3u-editor/commit/0bde7ad7ac397a176efc5156da2a5bd55e508332

PR #649, opened 2026-01-16
fix: use relationLoaded to avoid group column/relation conflict in channel API

Serph91P:
This pull request updates how group information is accessed in the `ChannelController` to avoid conflicts with a possible `group` column in the database. The changes ensure that Eloquent relationships are checked using `relationLoaded` and accessed via `getRelation` instead of direct property access.

**Improvements to relationship handling:**

* Updated group info construction in the `index` method to use `relationLoaded('group')` and `getRelation('group')` for safer and more explicit relationship access, preventing conflicts with a `group` column.
* Made the same update in the `show` method to consistently use `relationLoaded('group')` and `getRelation('group')` for group information.

### DIFF
--- a/app/Http/Controllers/ChannelController.php
+++ b/app/Http/Controllers/ChannelController.php
@@ -182,12 +182,12 @@ class ChannelController extends Controller
                     ];
                 }
 
-                // Build group info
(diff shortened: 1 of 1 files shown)


sparkison:
Good catch! Merging now 😉

PR #677, opened 2026-01-26
implement failover URL's for Xtream providers

djrarky:
implement failover URL's for Xtream providers

for both Playlist's and Playlist Aliases


sparkison:
I think understand what you're after here, but wondering how well it will work in reality. It looks like it's setup to determine the URL on sync only, so if a URL goes down in-between syncs, it won't do much other than add some additional overhead. And if we check mid-stream, or when a streaming URL is first requested it would add a lot of additional requests and delay stream start.

Or maybe I'm misunderstanding it? That's just how I'm seeing it, so wondering if it's really something we should add or not 🤔

djrarky:
Only on sync is ideal imo. 

I wasn't really aiming for instant failover and adding additional overhead there, there's already issues when you've got 1 stream and switch channels quickly, m3u doesn't close the old connection immediately. 

If sync fails, try the next domain, and if one works use that.

That's what I thought anyways. 
Sure you're restricted till the next sync, but you could run it manually, and it's very rare we see issues like we did last weekend.

sparkison:
> Only on sync is ideal imo.
> 
> I wasn't really aiming for instant failover and adding additional overhead there, there's already issues when you've got 1 stream and switch channels quickly, m3u doesn't close the old connection immediately.
> 
> If sync fails, try the next domain, and if one works use that.
> 
> That's what I thought anyways. Sure you're restricted till the next sync, but you could run it manually, and it's very rare we see issues like we did last weekend.

thanks for the update! That makes sense, I'll circle back and review again with that in mind...

sparkison:
After taking another look, I think it's just going to add confusion (what URLs get update, which don't, etc.) and would rather force a URL update directly if needed since it sounds like something that doesn't happen often. I can revisit if this comes up more often, but gonna close it down for now...

djrarky:
I think that's fair enough, my preference however would be for a single source of truth.

djrarky:
I think for that reason it should be considered, it'd be nice to not have to keep a spreadsheet with provider details etc.

Issue #669, opened 2026-01-23
.strm generation path for VOD Groups and Series Categories


bilo35x:
Hi guys,

first of all, I want to say thank you for such a masterpiece of sw. 

you made already a nice feature that allows besides global .strm settings, media based strm settings that can override the settings for each one of a vod or series.

It would be very nice if you take this to a wider view an allow this for vod groups and series categories. 

And the best case would be, if it's possible to not set that for each one categorie or group, then to set that for multiple groups or categories with just one input.

I hope you all doing well and i wish you the best.

thanks.

sparkison:
Feature added! Stream file sync settings have been moved to their own section and can be assigned now, rather than have various place to input the settings.

This feature is currently only available in `experimental` at the moment.

Head to Stream File Settings (under Playlists) to use the new feature:

<img width="1593" height="670" alt="Image" src="https://github.com/user-attachments/assets/3e36575b-d4ff-4488-9d4d-a6e03aec19c9" />

You can then assign these settings profiles to Groups, Categories, Series and VOD content, as well as within the Settings > Sync Settings

<img width="1586" height="516" alt="Image" src="https://github.com/user-attachments/assets/c0ff9fae-e9bc-4b6e-aad9-a3c166c08dea" />

bilo35x:
Nice! Good Job.

bilo35x:
When do you get this into release? Can you please put that into next release 0.33

sparkison:
> When do you get this into release? Can you please put that into next release 0.33

This is still in `experimental` for testing, so it will probably make it into `dev` in the coming days/weeks, and then into the main `release` branch usually a few weeks after. This is just so we can test these features more thoroughly and try to work through as many bugs as possible with smaller groups before rolling out to the masses 😉

PR #679, opened 2026-01-27
feat: Move Stream File sync settings to models for modular setup

sparkison:
Move the stream file sync settings from settings and model directly and allow creating a global Stream File Setting (SFS) and assigning to models - VOD/Series, Group/Category, and Settings (global).

This is the priority: Model (VOD/Series), if not set use Group/Category, if not set use Settings.


sparkison:
This PR resolves issue #669

PR #685, opened 2026-01-29
Feature/merge same id per group

Serph91P:
Add the ability to merge channels with the same stream ID within a
specific group, rather than across all playlists globally.

Changes:
- MergeChannels job: Add optional groupId parameter to filter channels
  by group when merging. When provided, only channels within that group
  are considered for merging.
- GroupResource: Add 'Merge Same ID for Group' action to the table row
  dropdown menu with full modal options (preferred playlist, failover
  playlists, resolution ordering, etc.)
- ViewGroup: Add the same action to the Actions dropdown on the group
  view page for easy access when viewing group details

This allows users to perform targeted merging operations on a per-group
basis, which is useful when working with large playlists where global
merging would be too broad

### DIFF
--- a/app/Jobs/UnmergeChannels.php
+++ b/app/Jobs/UnmergeChannels.php
@@ -78,9 +78,16 @@ class UnmergeChannels implements ShouldQueue
 
     protected function sendCompletionNotification()
     {
+        if ($this->playlistId) {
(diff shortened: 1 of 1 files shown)


sparkison:
Looks good! Just added the unmerge option too for consistency 😉

Issue #691, opened 2026-01-30
media server be able to choose libraries to sync


Serph91P:
**Is your feature request related to a problem? Please describe.**
You can reimport Data from the media server you exported as strm files.

**Describe the solution you'd like**
Be able to choose the the libraries to import to not import the same media again you exported from another playlist.

**Describe alternatives you've considered**

**Additional context**
Kinda funny that you can reimport the same media again.

Grimothy:
**Current Progress**
<img width="797" height="535" alt="Image" src="https://github.com/user-attachments/assets/50178082-97bd-47f1-8168-c591200c0072" />

Grimothy:
@sparkison  I think we can close this

PR #698, opened 2026-01-31
Feat: media server be able to choose libraries to sync #691

Grimothy:
## Whats changed

This PR is in relation to #691 691

It  allows users to specific what libraries to user with the media server integration


sparkison:
looks good, merging! Tested in Plex and Emby and confirmed working as expected 😉

PR #719, opened 2026-02-05
feat: Add inline custom group creation when adding items to custom playlists

Serph91P:
- Add toggle and text input to create new groups directly in 'Add to Custom Playlist' dialogs
- Implemented in ChannelResource, GroupResource, VodResource, VodGroupResource, CategoryResource, SeriesResource
- Also added to ViewGroup, ViewVodGroup, and ViewCategory header actions
- Groups are created as Spatie Tags with the custom playlist UUID as type
- Categories use UUID-category suffix for proper separation


sparkison:
Closing this down as I've updated the way we're doing this in `experimental` and would prefer not to try and merge this. Check out the latest `experimental` if you haven't! The actions have been moved to the `PlaylistService` to make it easier to update, and I've also added the option to use the existing model Group/Category to automatically create and assign them, options! 😉

PR #728, opened 2026-02-10
Implement alias expiration validation

Aldione:
Add expiration check for playlist alias


sparkison:
Had a couple changes that this was conflicting with, so implemented directly 😉

PR #732, opened 2026-02-12
Implement user permissions

sparkison:
Add simple user permissions to restrict what user accounts can access.


sparkison:
Added stream file sync and fixed the tools permission assignment

permissions

PR #740, opened 2026-02-16
feat: add popout player tab from floating player

pablopunk:
First of all thank you very much for this repo, it has replaced a few things I've tried that were always a nightmare to set up and this one solves it all beautifully 🩵.

I was wondering if we are interested in a full-size player? i.e. you can open the small player on a new tab. I know you can already play the video fullscreen, but this is more useful/handy IMO:

<p>CleanShot 2026-02-16 at 14 45 43</p>
<p>CleanShot 2026-02-16 at 14 45 51</p>
<p>CleanShot 2026-02-16 at 15 19 20</p>

### DIFF
--- a/app/Http/Controllers/PlayerController.php
+++ b/app/Http/Controllers/PlayerController.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+
+class PlayerController extends Controller
+{
+    /**
+     * Display the popout player view.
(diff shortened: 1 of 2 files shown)


sparkison:
I've updated it a bit as the overlay didn't align with the floating player, and when opening a new tab, the floating player remained active. Now it will close the floating player when opening in a new window and the info/overlay aligns with the floating player.

Thanks for the implementation! I think this is a great feature 😉

PR #776, opened 2026-02-27
Fix TMDB metadata preservation during sync

Grimothy:
## Summary
Fix critical bug where TMDB metadata was being wiped on every sync cycle for local media integrations, causing full re-fetches of all 348 movies and 109 series from TMDB every time.

### Changes Made
- **SyncMediaServer.php**: Replaced `updateOrCreate` with `firstOrNew` + conditional merge logic to preserve TMDB-populated fields (plot, cover, cast, etc.) when sync data is empty or local paths.
- **ListVod.php**: Update manual TMDB search to populate dedicated `tmdb_id`/`imdb_id` columns and clear `last_metadata_fetch` for full re-fetch.
- Prevented logo/logo_internal from being overwritten with empty values on sync.

### Impact
- Sync cycles now complete in ~3 seconds instead of hours
- TMDB API calls reduced from ~457 per sync to ~1 (only for incomplete metadata)
- Posters and metadata preserved across syncs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -858,9 +858,9 @@
             }
         },
         "node_modules/@rollup/rollup-android-arm-eabi": {
-            "version": "4.55.1",
-            "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.55.1.tgz",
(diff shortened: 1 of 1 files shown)


Grimothy:
Closing to sync branches before merging TMDB fixes

PR #802, opened 2026-03-08
fix: invalidate EPG XML cache when channel mappings change

gabelul:
## The Problem

When EPG channel mappings are added (auto-mapping) or removed (bulk unmap), the cached XMLTV files aren't invalidated. The `/{uuid}/epg.xml` endpoint caches generated XML for 6 hours, so Xtream API clients (TVMate, TiviMate, etc.) keep serving stale data that may be missing newly mapped channels entirely.

The `clearPlaylistEpgCacheFile()` method and `Epg::getAllPlaylists()` already exist and work great — they're called for M3U imports, EPG cache regeneration, and playlist setting changes. They just weren't wired up for the most common EPG change: mapping channels to EPG sources.

## The Fix

| File | Change |
|------|--------|
| `MapEpgToChannelsComplete.php` | Clear EPG cache for all affected playlists when auto-mapping completes (uses existing `$epg->getAllPlaylists()` pattern from `GenerateEpgCache.php`) |
| `ChannelResource.php` | Clear EPG cache for affected playlists (regular, custom, merged) when channels are bulk-unmapped |

## How I Found It

Custom playlist with 262 EPG-mapped channels was serving a stale XMLTV with only 18 channels (from before most mappings were set up). TVMate showed "no information" for every channel. Deleting the cache file manually fixed it immediately — which confirmed the cache invalidation gap.

## Test Results

Verified on a live instance with a custom playlist:

| Test | Before Fix | After Fix |
|------|-----------|-----------|
| XMLTV channels | 18 (stale cache) | 262 |
| XMLTV programmes | 676 | 11,270 |
| `get_short_epg` working | Yes (per-channel, no cache) | Yes |
| `epg_channel_id` ↔ XMLTV match | 18/262 | 262/262 |
| Cache cleared on mapping | No | Yes |

### DIFF
--- a/app/Filament/Resources/Channels/ChannelResource.php
+++ b/app/Filament/Resources/Channels/ChannelResource.php
@@ -734,21 +734,18 @@ class ChannelResource extends Resource
                     ->modalSubmitActionLabel('Map now'),
                 BulkAction::make('unmap')
                     ->label('Undo EPG Map')
-                    ->action(function (Collection $records, array $data): void {
-                        $channelIds = $records->pluck('id')->toArray();
-
+                    ->action(function (Collection $records): void {
                         // Clear the EPG mapping
-                        Channel::whereIn('id', $channelIds)
+                        Channel::whereIn('id', $records->pluck('id'))
                             ->update(['epg_channel_id' => null]);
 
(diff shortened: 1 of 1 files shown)


sparkison:
great catch! Thank you for the contribution 🙏

PR #810, opened 2026-03-11
feat: Show provider profile name on Stream Monitor page

Warbs816:
Displays which provider profile each stream is using as a badge next to the stream format. Shows "Provider Profile: Primary" for the primary account or "Provider Profile: [Name]" for additional profiles.


sparkison:
Thanks for the contribution! Merging 😉

PR #811, opened 2026-03-12
Provider profile fixes: capacity enforcement, bypass toggle, and client affinity

Warbs816:
## Summary

- **Re-enable `available_streams` enforcement** when Provider Profiles are enabled, and add a `bypass_provider_limits` toggle so users can allow streams to start even when all provider profiles are at capacity (force-selects the least-loaded profile)
- **Client-to-profile affinity via Redis** — when a client is assigned a provider profile, the mapping is stored so subsequent requests from the same client prefer the same profile. This prevents consuming an extra connection slot when switching channels before the old stream's decrement webhook fires
- **Fix PlaylistInfo profile display** to use actual proxy counts instead of stale Redis counts

### Bypass Provider Limits toggle
- New `bypass_provider_limits` boolean column on playlists
- `selectProfile()` gains a `forceSelect` mode that picks the least-loaded profile when all are at capacity
- UI toggle added to PlaylistResource

### Client-to-Profile Affinity
- Client identity: `{ip}:{username}` (falls back to IP-only when username unavailable)
- Redis key: `client_affinity:{clientIdentifier}:{playlistId}` → `{profileId}`
- 24-hour TTL, refreshed on each use
- Always-on when profiles are enabled — no new toggle needed
- Affinity profile is always preferred (even if it appears at capacity, since the client's old stream likely hasn't been decremented yet)

### Files changed
- `ProfileService.php` — affinity helpers, `selectProfile()` affinity check, `selectAndReserveProfile()` stores affinity
- `M3uProxyService.php` — builds client identifier and passes it through all profile selection calls
- `M3uProxyApiController.php` — passes `$request` to `getEpisodeUrl()` for IP extraction
- `PlaylistResource.php` / `PlaylistInfo.php` / `Playlist.php` — bypass toggle UI and model
- Migration for `bypass_provider_limits` column

### DIFF
--- a/app/Filament/GuestPanel/Resources/Vods/VodResource.php
+++ b/app/Filament/GuestPanel/Resources/Vods/VodResource.php
@@ -193,7 +193,9 @@ class VodResource extends Resource
                     ->label('Default URL')
                     ->sortable()
                     ->searchable(query: function (Builder $query, string $search): Builder {
-                        return $query->orWhereRaw('LOWER(channels.url::text) LIKE ?', ['%'.strtolower($search).'%']);
+                        $urlExpr = DB::getDriverName() === 'sqlite' ? 'channels.url' : 'channels.url::text';
+
+                        return $query->orWhereRaw("LOWER({$urlExpr}) LIKE ?", ['%'.strtolower($search).'%']);
                     })
                     ->toggleable(isToggledHiddenByDefault: true),
 
--- a/app/Filament/Pages/M3uProxyStreamMonitor.php
+++ b/app/Filament/Pages/M3uProxyStreamMonitor.php
@@ -6,6 +6,7 @@ use App\Facades\LogoFacade;
 use App\Models\Channel;
 use App\Models\Episode;
 use App\Models\Network;
+use App\Models\PlaylistProfile;
 use App\Models\StreamProfile;
 use App\Services\M3uProxyService;
 use Carbon\Carbon;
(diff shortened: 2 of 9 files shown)


Grimothy:
## Query Storm Resolution Confirmation

This PR resolves the critical query storm issue reported in m3u-proxy logs, where provider profiles at capacity were triggering hundreds of `/streams/by-metadata` queries per second.

### Root Cause Analysis:

The issue was introduced in two commits:
1. **Commit `0bcb12df`** (March 12, by @Warbs816) - attempted to fix VOD/movie 503 failures by adding `$forceSelect` and `$clientIdentifier` parameters, but these didn't exist yet in the function signature
2. **Commit `cf7a63f1`** (March 13, by Shaun Parkison) - fixed the PHP errors by removing the non-existent parameters, but accidentally removed ALL parameters including the critical `$channelId` and `$channelPlaylistUuid` needed for reuse detection

### How This PR Fixes It:

✅ Restores the missing `$id` and `$playlist->uuid` parameters to `selectAndReserveProfile()` calls in `getEpisodeUrl()` (lines 1131, 1169, 1177)

✅ Properly implements the `$forceSelect` and `$clientIdentifier` parameters in the function signature

✅ Adds reuse detection logic to prevent allocating new slots when streams already exist

✅ Includes comprehensive tests (537 lines) to prevent regression

### Additional Enhancements:

- **Client affinity**: Prevents double slot allocation during rapid channel switches
- **Bypass provider limits toggle**: Allows streaming when provider reports at capacity
- **Force-select mode**: Picks least-loaded profile when all profiles at capacity

### Status:

This PR is ready for review and merge. It provides a complete solution to the query storm issue plus valuable additional features for provider profile management. 🚀

PR #813, opened 2026-03-12
Fix VOD/movie 503 failures with Provider Profiles enabled

Warbs816:
## Summary

- **Fixed `findExistingPooledStream()` only matching transcoded streams** — when reuse detection fired for direct streams (movies, timeshift, VOD without a transcoding profile), it couldn't find the existing stream and fell through to 503. Now correctly matches direct (non-transcoded) streams when `profileId` is null.
- **Fixed `getEpisodeUrl()` missing `channelId`/`channelPlaylistUuid` params** in all three `selectAndReserveProfile()` calls, preventing atomic reuse detection for episodes.
- **Added reuse detection handling to `getEpisodeUrl()`** — when `selectAndReserveProfile()` returned `[null, null]` due to channel reuse, it went straight to "no capacity" instead of joining the existing stream. Now has the same reuse detection + existing stream lookup that `getChannelUrl()` has.
- Added diagnostic logging to `transformUrl()` and direct stream creation to aid future debugging.

### DIFF
--- a/app/Filament/GuestPanel/Resources/Vods/VodResource.php
+++ b/app/Filament/GuestPanel/Resources/Vods/VodResource.php
@@ -194,6 +194,7 @@ class VodResource extends Resource
                     ->sortable()
                     ->searchable(query: function (Builder $query, string $search): Builder {
                         $urlExpr = DB::getDriverName() === 'sqlite' ? 'channels.url' : 'channels.url::text';
+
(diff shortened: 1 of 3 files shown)


sparkison:
Thanks for this! I've pulled out some the of the items from the other PR and will merge this back in there and re-add.

PR #824, opened 2026-03-14
Broadcasting fix

Grimothy:
Addresses issues with hls variables not being passed

### DIFF
--- a/docker-compose.embedded-redis.yml
+++ b/docker-compose.embedded-redis.yml
@@ -83,18 +83,6 @@ services:
       # Note: ROOT_PATH=/m3u-proxy is now the default, no need to set it explicitly
       # Only set ROOT_PATH= (empty) if you need to use the root path instead
 
-      # HLS Segment Storage & Garbage Collection
-      - HLS_TEMP_DIR=${HLS_TEMP_DIR:-/var/www/html/storage/app/hls-segments}
-      - HLS_GC_ENABLED=${HLS_GC_ENABLED:-true}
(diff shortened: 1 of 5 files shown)


sparkison:
Still having a couple issues with this one:

1. Remove `output_dir` from the broadcast payload (use `HLS_BROADCAST_DIR`)
2. Rename `BROADCAST_TEMP_DIR` → passes as `HLS_BROADCAST_DIR` to the proxy in supervisord/docker-compose
3. Drop `BROADCAST_GC_INTERVAL` / `BROADCAST_GC_AGE_THRESHOLD` entirely — they're ignored by the proxy; `HLS_GC_INTERVAL` / `HLS_GC_AGE_THRESHOLD` already cover this
4. Replace `.local-proxy/` with a configurable build arg

Grimothy:
I suppose we can close it. I would reference the discord chat. I think we have a configuration deviation from embedded vs proxy.

The temp directory for broadcast needs additional parameters

Grimothy:
Closed reference the proxy PR

sparkison:
Sorry, wasn't saying close it! Just needed some edits 😉 ...I see what you're after, and think I got it!

To use a local copy of the proxy (for testing AIO), use `M3U_PROXY_LOCAL_DIR` in your docker build (`.env.docker`) and set it relative to the editor. E.g. if it's in an adjacent directory, it would be `M3U_PROXY_LOCAL_DIR=../m3u-proxy`. Wanted to make that one a little more configurable/universal.

PR #826, opened 2026-03-14
fix: Client affinity independent of bypass_provider_limits

### LINKED ISSUE #825: Client-to-profile affinity only works when Bypass Provider Connection Limits is enabled
## Description

Client-to-profile affinity (introduced in PR #811, commit 00314f2) does not work unless the **Bypass Provider Connection Limits** toggle is enabled on the playlist.

## Expected behavior

When Provider Profiles are enabled, client affinity should always prefer the previously-assigned profile regardless of the bypass toggle. The whole purpose of affinity is to handle the channel-switch window where the old stream's decrement webhook hasn't fired yet — the profile *appears* full but the client already occupies one of those slots.

## Actual behavior

Affinity only bypasses capacity when `bypass_provider_limits` is `true`. With it off, the affinity profile is skipped whenever it appears at capacity, falling back to normal priority-based selection. This causes unnecessary profile switches and can trigger "no profiles with capacity" 503 errors during channel changes.

## Root cause

In `ProfileService::selectProfile()`, the affinity check is gated behind `$forceSelect`:

```php
if ($affinityProfile && ($forceSelect || static::hasCapacity($affinityProfile))) {
```

`$forceSelect` is set from `$playlist->bypass_provider_limits`, so affinity is effectively disabled without that toggle.

This was introduced in commit 40ba277 which rolled back the original always-prefer design from commit edf0725. The concern was overcounting, but the temporary over-count during channel switches is by design and self-corrects when the old stream decrements. Channel reuse detection in `selectAndReserveProfile` already prevents double-allocation for the same channel.

## Fix

Remove the `$forceSelect || hasCapacity` guard — always return the affinity profile when one exists.

Warbs816:
## Summary

- **Fixes #825** — Client-to-profile affinity was only working when the "Bypass Provider Connection Limits" toggle was enabled
- Removes the `$forceSelect || hasCapacity` guard on the affinity path in `selectProfile()` so affinity always prefers the previously-assigned profile
- Updates tests to reflect the corrected always-prefer behavior

## Detail

The affinity capacity check was gated behind `$forceSelect` (which maps to `bypass_provider_limits`), so affinity only bypassed capacity with that toggle on. This defeats the purpose of affinity — during channel switches the old stream's decrement webhook hasn't fired yet, so the profile *appears* full but the client already occupies one of those slots.

Channel reuse detection in `selectAndReserveProfile` already prevents double-allocation for the same channel, and the temporary over-count during switches self-corrects when the old stream decrements.

### DIFF
--- a/app/Services/ProfileService.php
+++ b/app/Services/ProfileService.php
@@ -146,7 +146,7 @@ class ProfileService
             if ($affinityProfileId !== null) {
                 $affinityProfile = $profiles->firstWhere('id', $affinityProfileId);
 
-                if ($affinityProfile && ($forceSelect || static::hasCapacity($affinityProfile))) {
+                if ($affinityProfile) {
                     Log::debug('Returning affinity profile for client', [
                         'client_identifier' => $clientIdentifier,
                         'profile_id' => $affinityProfile->id,
--- a/tests/Feature/ClientProfileAffinityTest.php
+++ b/tests/Feature/ClientProfileAffinityTest.php
@@ -129,12 +129,13 @@ test('selectProfile reuses affinity profile when it has capacity', function () {
         ->and($selected->id)->toBe($secondProfile->id);
 });
 
-// ── Affinity to at-capacity profile → falls back to next available ────────
+// ── Affinity to at-capacity profile → still uses affinity ─────────────────
+// The client's old stream likely hasn't been decremented yet, so the profile
+// appears full but the client already occupies one of those slots.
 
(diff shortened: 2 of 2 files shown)


sparkison:
I'm going to introduce a toggle for this with default `false` so you can enable affinity per-playlist, with the default behavior for existing installs to not utilize this feature. It will work like the **Bypass Provider Connection Limits** toggle in that regard 😉

Warbs816:
Even better! Thank you

PR #828, opened 2026-03-14
fix: 500 error when creating a new Playlist Alias

### LINKED ISSUE #827: 500 error when clicking New Playlist Alias button
## Description

Clicking the "New Playlist Alias" button on the playlist-aliases page throws a 500 error:

```
ErrorException
Attempt to read property "playlist_id" on null
app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php:416
```

## Root cause

The `hidden()` callback on the "Get from playlist status" hint action accesses `$record->playlist_id`, but `$record` is `null` during creation (there is no existing record yet):

```php
->hidden(fn ($record) => $record->playlist_id === null)
```

## Fix

Use the null-safe operator: `$record?->playlist_id`

When `$record` is `null`, this evaluates to `null === null` → `true`, keeping the button hidden during creation — which is correct since there's no playlist to detect timezone from yet.

Warbs816:
## Summary

- **Fixes #827** — Clicking "New Playlist Alias" threw a 500 `ErrorException: Attempt to read property "playlist_id" on null`
- Uses null-safe operator (`$record?->playlist_id`) in the `hidden()` callback on the timezone hint action, since `$record` is `null` during creation

### DIFF
--- a/app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php
+++ b/app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php
@@ -413,7 +413,7 @@ class PlaylistAliasResource extends Resource
                                         ->body('Provider timezone not found in playlist status. Make sure the playlist is connected and has synced at least once to retrieve this information.')
                                         ->danger()
                                         ->send();
-                                })->hidden(fn ($record) => $record->playlist_id === null)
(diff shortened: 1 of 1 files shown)


sparkison:
Love when I miss things so simple 🤦

PR #838, opened 2026-03-17
fix: Enforce capacity check on affinity profile selection

### LINKED ISSUE #837: Client-provider affinity bypasses capacity check, exceeding provider max connections
## Bug Description

When client-provider affinity is enabled, `selectProfile()` returns the affinity profile **without checking whether it has available capacity**. This causes the provider's `max_streams` limit to be exceeded when a pooled stream outlives the client that created it.

### Scenario

1. **Device A** watches Channel 1 → assigned to **Provider A** (affinity stored)
2. **Device B** watches Channel 1 → pools from Provider A's upstream connection via the proxy's broadcast system (no new upstream connection opened)
3. **Device A** switches to Channel 2 → affinity returns **Provider A** again (no capacity check)
4. Provider A now has **2 upstream connections** (Channel 1 for Device B + Channel 2 for Device A) but `max_streams = 1`
5. Provider bans the account for exceeding connection limits

### Root Cause

In `ProfileService::selectProfile()`, the affinity code path (lines 150-157) returns the affinity profile immediately without calling `hasCapacity()`. The normal priority-based selection path does check capacity, but affinity bypasses it entirely.

Additionally, `hasCapacity()` relies solely on the Redis connection counter, which can drift when pooled streams outlive their original client. When Device A disconnects from Channel 1, the webhook decrements the Redis counter to 0, but Device B still holds the upstream connection — the proxy has 1 active stream but Redis says 0.

### Additional Issue

`reconcileFromProxy()` uses `client_count` (proxy-level clients) instead of counting stream entries (upstream connections) when building the profile stream count map. This over-reports when multiple clients pool from a single upstream connection.

## Expected Behavior

- Affinity should respect provider capacity — if the affinity profile is at `max_streams`, fall through to the next available provider
- `hasCapacity()` should cross-check with the proxy API for actual upstream connection count, not just the Redis counter
- `reconcileFromProxy()` should count streams (1 per upstream connection), not clients

Warbs816:
## Summary

Fixes #837 — client-provider affinity now respects provider capacity limits instead of blindly returning the affinity profile.

### Changes

**`app/Services/ProfileService.php`**

- **Add capacity check to affinity path** — `selectProfile()` now calls `hasCapacity()` before returning the affinity profile. If the affinity profile is at capacity, it logs a message and falls through to normal priority-based selection, routing the client to a different provider.

- **Proxy-aware capacity check** — `hasCapacity()` now queries the proxy API for the actual upstream connection count (`getActiveStreamsCountByMetadata`) and takes `max(proxyCount, redisCount)`. The Redis counter can drift when pooled streams outlive their original client (subscriber inherits the upstream connection after the primary disconnects, but the Redis counter was already decremented). The proxy count reflects real upstream connections.

- **Fix reconcileFromProxy() stream counting** — Each stream entry represents one upstream provider connection, regardless of how many proxy clients are pooling from it. Changed from `+= client_count` to `+= 1`.

**Tests**

- Updated `ClientProfileAffinityTest` — the "affinity at capacity" test now asserts the correct behavior (falls back to next available profile instead of returning the at-capacity affinity profile)
- Added `Http::fake` to `BypassProviderLimitsTest`, `ProfileAuditFixesTest`, and `ProfileConnectionRaceConditionTest` since `hasCapacity()` now makes a proxy API call

### Testing

- ✅ Affinity profile returned when it has capacity (existing behavior preserved)
- ✅ Affinity profile skipped when at capacity → falls back to priority-based selection
- ✅ Capacity check sees real upstream connections via proxy API (not just Redis counter)
- ✅ All existing tests updated and passing

### DIFF
--- a/app/Services/ProfileService.php
+++ b/app/Services/ProfileService.php
@@ -147,7 +147,7 @@ class ProfileService
             if ($affinityProfileId !== null) {
                 $affinityProfile = $profiles->firstWhere('id', $affinityProfileId);
 
-                if ($affinityProfile) {
+                if ($affinityProfile && static::hasCapacity($affinityProfile)) {
                     Log::debug('Returning affinity profile for client', [
                         'client_identifier' => $clientIdentifier,
                         'profile_id' => $affinityProfile->id,
@@ -155,6 +155,16 @@ class ProfileService
                     ]);
 
                     return $affinityProfile;
+                }
+
+                if ($affinityProfile) {
+                    Log::info('Affinity profile at capacity, selecting next available', [
+                        'client_identifier' => $clientIdentifier,
+                        'affinity_profile_id' => $affinityProfile->id,
+                        'affinity_profile_name' => $affinityProfile->name,
+                        'active_connections' => static::getConnectionCount($affinityProfile),
+                        'max_connections' => $affinityProfile->effective_max_streams,
+                    ]);
                 }
             }
         }
@@ -406,6 +416,10 @@ class ProfileService
 
     /**
      * Check if a profile has available capacity.
+     *
+     * Uses the proxy API as the source of truth for upstream connection
+     * count (not the Redis counter, which can drift when pooled streams
+     * outlive the client that originally created them).
      */
     public static function hasCapacity(PlaylistProfile $profile): bool
     {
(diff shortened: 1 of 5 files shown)


sparkison:
Implemented!

PR #850, opened 2026-03-19
fix: don't demand a CSRF token for group API endpoint

djrarky:


sparkison:
Good catch! Thanks for this 😉

PR #851, opened 2026-03-19
feat: Change preferred icon source on EPG map

djrarky:
New toggle in the advanced settings of the EPG map modal, allowing if preferred the option to set the icon source of channels to be from the EPG.


sparkison (approved):
Works for me! Approved and merging 😉

PR #853, opened 2026-03-20
fix: Resolve channel icons not displaying in EPG playlist viewer

Warbs816:
## Summary
- Channel icons in the EPG playlist viewer always showed the placeholder image because the icon resolution code referenced an undefined variable (`$epgData`)
- The null coalescing operator silently swallowed the error, causing every EPG-type channel to fall through to `placeholder.png`
- Fixed by using the already-joined `epg_channels` fields (`$channel->epg_icon`, `$channel->epg_icon_custom`), consistent with the EPG channel map block and Xtream API controller

## Test plan
- [ ] Open a playlist overview page with the EPG viewer visible
- [ ] Verify channel logos appear next to channel names (for both EPG and Channel logo types)
- [ ] Verify logo proxy works correctly when enabled
- [ ] Verify channels without any logo still show the placeholder

### DIFF
--- a/app/Http/Controllers/Api/EpgApiController.php
+++ b/app/Http/Controllers/Api/EpgApiController.php
@@ -351,10 +351,11 @@ class EpgApiController extends Controller
                 if ($channel->logo) {
                     // Logo override takes precedence
                     $icon = $channel->logo;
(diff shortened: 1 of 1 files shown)


Warbs816:
Targeted wrong branch and accidentally opened a dupe instead of switching branch, sorry!

PR #855, opened 2026-03-20
feat: add VPN watchdog UI to Stream Monitor

Serph91P:
- Add fetchVpnStatus() and triggerVpnRotation() to M3uProxyService
- Add VPN status card to Stream Monitor (state, IP, country, latency)
- Add manual VPN rotation button with confirmation
- Add VPN webhook event handling in ApiController

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,6 @@ APP_NAME="m3u editor"
 APP_ENV=production
 APP_KEY=
 APP_DEBUG=false
-APP_TIMEZONE=UTC
 APP_URL=http://localhost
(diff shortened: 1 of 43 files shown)


Serph91P:
@sparkison could you take a look at this PR as well as the related one linked here: https://github.com/m3ue/m3u-proxy/pull/46? Both are connected and it would be great if you could review them together.

sparkison:
This feels separate and distinct from the editor and something that would be better suited as it's own app or Docker container. Closing it down!

PR #860, opened 2026-03-23
Fix HLS stream detection for URLs with query parameters

### LINKED ISSUE #859: HLS streams with query parameters served as direct streams instead of being proxied
## Bug Description

M3U8 URLs that contain query parameters (e.g. `channel.m3u8?location=ZEOP01`) are incorrectly detected as direct streams instead of HLS streams. This causes the proxy to dump the raw playlist file (~917 bytes) to the client instead of processing it through the HLS playlist rewriter.

## Root Cause

`pathinfo()` in PHP does not understand URL query strings. For a URL like `antreunihd.m3u8?location=ZEOP01`, `pathinfo($url, PATHINFO_EXTENSION)` returns `m3u8?location=ZEOP01` instead of `m3u8`. This fails the `=== 'm3u8'` check in `M3uProxyService`, so `buildProxyUrl()` routes to `/stream/{id}` (direct) instead of `/hls/{id}/playlist.m3u8`.

## Steps to Reproduce

1. Add a channel with an M3U8 URL that includes query parameters (e.g. `https://example.com/stream.m3u8?token=abc`)
2. Enable proxy and attempt to play the channel
3. The stream does not play — the m3u-proxy logs show a single 917-byte chunk delivered as a direct stream

## Expected Behavior

The stream should be detected as HLS and routed through the HLS playlist rewriter, allowing proper playback.

## Affected Code

5 instances of `pathinfo($url, PATHINFO_EXTENSION)` in `app/Services/M3uProxyService.php` (lines 794, 918, 1118, 1233, 1374).

Warbs816:
## Summary

- Strip query strings from URLs before extracting file extension via `pathinfo()`, using `parse_url($url, PHP_URL_PATH)` 
- Fixes all 5 instances in `M3uProxyService.php` where format detection was broken for URLs with query parameters

Closes #859

### DIFF
--- a/app/Services/M3uProxyService.php
+++ b/app/Services/M3uProxyService.php
@@ -791,7 +791,7 @@ class M3uProxyService
                 ]);
 
                 $url = PlaylistUrlService::getChannelUrl($channel, $playlist);
-                $format = pathinfo($url, PATHINFO_EXTENSION);
+                $format = pathinfo(parse_url($url, PHP_URL_PATH) ?? $url, PATHINFO_EXTENSION);
                 $format = $format === 'm3u8' ? 'hls' : $format;
 
                 return $this->buildProxyUrl($existingStreamId, $format, $username);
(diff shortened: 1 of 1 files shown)


sparkison:
Closing as already implemented 😉 ...still need to get it pushed...

PR #861, opened 2026-03-23
Fix 500 errors in EpgViewer and getFloatingPlayerAttributes

Warbs816:
## Summary

- Add null-safe access for playlist relationship in `EpgViewer.php` — prevents crash when a channel's parent playlist is missing
- Add default `null` values for `$username` and `$password` params in `Channel::getFloatingPlayerAttributes()` and `Episode::getFloatingPlayerAttributes()` — most callers (admin panel play buttons) don't pass credentials


sparkison:
Appreciate it! 🙏

PR #868, opened 2026-03-24
feat: Add channel scrubber to check Live (and optionally VOD) streams via http or ffprobe

sparkison:
## Channel Scrubber

A new **Channel Scrubber** feature has been added under the **Playlist** navigation group. It automatically identifies and disables channels with broken or invalid stream links.

---

### Overview

Each scrubber is tied to a specific playlist and can be run manually on demand or triggered automatically after each playlist sync. Results are logged per run, retaining up to 15 entries.

---

### Configuration

| Option | Description |
|---|---|
| **Playlist** | The playlist whose channels will be checked |
| **Check Method** | `HTTP` (fast, HEAD request) or `FFprobe` (slower, more reliable) |
| **Include VOD** | Optionally extend checks to VOD channels (live channels are always checked) |
| **Recurring** | Auto-run after each playlist sync |
| **Scan All Channels** | By default only enabled channels are checked; toggle to also scan disabled ones (carries a warning about connection volume and run time) |

---

### Check Methods

- **HTTP** — Sends a HEAD request to each stream URL. Very fast, recommended for large playlists. May produce false positives for providers that reject HEAD requests.
- **FFprobe** — Attempts to open each stream and verify it returns valid media data. Significantly more reliable but much slower — expect long run times for large playlists.

---

### Lifecycle

- **On Create** — The scrubber runs immediately after being saved.
- **Manual Run** — "Run Now" action available on any idle scrubber.
- **Restart** — Abandons the current run and starts a fresh one.
- **Cancel** — Stops the in-progress run. Any chunks already dispatched will complete their current batch before halting. Channels already disabled during the run remain disabled.
- **Recurring** — When enabled, fires automatically at the end of each successful playlist sync (same trigger point as EPG maps).

---

### Run Logs

Each completed (or cancelled) run produces a log entry attached to the scrubber, accessible from the scrubber's detail view. Logs include:

- Run timestamp & status
- Channels checked
- Dead links found
- Channels disabled
- Runtime

A maximum of **15 log entries** are retained per scrubber; older entries are automatically pruned.

---

### Cancellation Mechanism

Cancellation works via two independent guards in the queued chunk and completion jobs:

1. **UUID mismatch** — each run is tagged with a batch UUID; a Restart generates a new UUID, causing stale queued jobs to bail immediately.
2. **Status check** — Cancel sets `status = cancelled`; jobs additionally check this before processing, providing a clean stop signal independent of restarts.

### DIFF
--- a/app/Filament/Resources/ChannelScrubbers/RelationManagers/ScrubberLogsRelationManager.php
+++ b/app/Filament/Resources/ChannelScrubbers/RelationManagers/ScrubberLogsRelationManager.php
@@ -41,7 +41,12 @@ class ScrubberLogsRelationManager extends RelationManager
                     ->toggleable(),
                 TextColumn::make('status')
                     ->badge()
-                    ->color(fn (string $state) => $state === 'completed' ? 'success' : 'danger')
+                    ->color(fn (string $state) => match ($state) {
+                        'completed' => 'success',
+                        'processing' => 'warning',
+                        'cancelled' => 'gray',
+                        default => 'danger',
+                    })
                     ->sortable()
                     ->toggleable(),
                 TextColumn::make('channel_count')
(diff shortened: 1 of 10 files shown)


sparkison:
Adding a preview, cause why not 🤷 

channel-scrubber

PR #871, opened 2026-03-24
Add option to use original source URL in .strm files

### LINKED ISSUE #521: Option to generate .strm files using original target URL
I use m3u-editor to create playlists for Channels DVR for live channel content. For VOD and Series I see m3u-editor supports generation of .strm files which in theory would be a killer addition for me since Channels DVR allows for import of Movies and TV via .strm file. However, in practice this does not work because Channels itself does not proxy the streaming via URL in the .strm files, it simply passes the exact URL over to the client which in my case is 'localhost' as I have m3u-editor and Channels running on the same host and do not wish to expose m3u-editor externally for simplicity and security.

Ideally under 'Sync Options' for each type of .strm file generation there would be an option like "Use original target URL" so that .strm files are created with the original media target which is very useful for those of us using clients like Channels DVR which just pass the .strm URL directly to the clients. The alternative here would be changing the APP_URL for m3u-editor to the LAN IP of the m3u-editor host which might solve the issue for devices on the local network, but this would not solve the issue when devices connect to Channels DVR remotely. For that an external FQDN would be needed which exposed m3u-editor through a VPN or reverse proxy which further increases the complexity and maintenance of such a setup.

Warbs816:
## Summary

- Adds a "URL Type" setting to Stream File Setting profiles: **Proxy URL** (default, existing behaviour) or **Original Source URL**
- When set to original, .strm files contain the direct provider URL instead of the m3u-editor proxy URL
- Applies to both VOD channels and series episodes
- Legacy settings default to proxy for backwards compatibility

Closes #521

## Test plan

- [x] Migration adds `url_type` column to `stream_file_settings`
- [x] Setting "Original Source URL" on a profile writes provider URLs to .strm files
- [x] Setting "Proxy URL" retains existing behaviour
- [x] CI passes

### DIFF
--- a/app/Filament/Resources/StreamFileSettings/StreamFileSettingResource.php
+++ b/app/Filament/Resources/StreamFileSettings/StreamFileSettingResource.php
@@ -100,13 +100,17 @@ class StreamFileSettingResource extends Resource
                 Select::make('url_type')
                     ->label('URL Type')
                     ->options([
-                        'proxy' => 'Proxy URL (m3u-editor)',
+                        'proxy' => 'M3U Editor (default)',
                         'original' => 'Original Source URL',
                     ])
                     ->default('proxy')
+                    ->native(false)
                     ->columnSpanFull()
-                    ->visible(fn (Get $get): bool => (bool) $get('enabled'))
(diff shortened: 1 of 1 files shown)


fabien4444:
+1 for this feature 👍

PR #878, opened 2026-03-26
fix: output set channel numbers correctly

djrarky:
When preferred output is "Channel Number", the code uses $channelNumber which is the auto-increment counter, not the channels actual number


sparkison:
Oh man, can't believe I missed this one 🤦 thank you!

PR #881, opened 2026-03-26
fix: Remove icon from Tools nav group to resolve Filament v4 conflict

Warbs816:
## Summary
- Filament v4 disallows both a navigation group and its items having icons simultaneously
- The `Tools` group had an icon set, while the `Backups` page inherited one from its vendor parent class, causing a 500 error (particularly visible with top navigation)
- Removed the group-level icon from `Tools` and properly nulled the `Backups` page icon override

## Test plan
- [x] Verified app loads without 500 error using top navigation
- [ ] Confirm Tools nav group renders correctly in both top and side nav modes
- [ ] Confirm all Tools group items (API Tokens, Assets, Backup & Restore, Post Processing, Debug Logs, Release Logs) are accessible

### DIFF
--- a/app/Filament/Pages/Backups.php
+++ b/app/Filament/Pages/Backups.php
@@ -16,7 +16,7 @@ use ShuvroRoy\FilamentSpatieLaravelBackup\Pages\Backups as BaseBackups;
 
 class Backups extends BaseBackups
 {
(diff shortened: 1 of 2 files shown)


Warbs816:
Seen that this is already resolved. Thanks

PR #882, opened 2026-03-26
fix: prevent browser validation errors on Settings page

gabelul:
## Summary

- Settings page throws "An invalid form control is not focusable" when saving, especially in Firefox
- Two root causes: collapsed sections with numeric inputs, and unconditionally required fields behind conditional toggles
- Adds `novalidate` to the form (Livewire validates server-side anyway) and makes `required()` conditional on parent toggle state

## What changed

**`app/Filament/Pages/Preferences.php`**

1. Override `getFormContentComponent()` to add `novalidate` attribute to the `<form>` element — prevents browsers from validating invisible fields inside collapsed sections (MediaFlow Proxy section has a `->numeric()` field in a `->collapsed(true)` section)

2. Make `provider_request_delay_ms` and `provider_max_concurrent_requests` conditionally required based on `enable_provider_request_delay` toggle, matching the existing `->hidden()` condition

## Reproduction

1. Open Settings page in Firefox
2. Leave "Provider Request Delay" toggle OFF and "MediaFlow Proxy" section collapsed
3. Change any setting and click Save
4. Browser blocks submission with "An invalid form control is not focusable"

Chrome is more lenient about validating hidden fields, which is why this mostly surfaces in Firefox.

## Test plan

- [v] Open Settings in Firefox with Provider Request Delay OFF — save should work
- [v] Toggle Provider Request Delay ON, leave delay/concurrent fields empty — server-side validation should catch it
- [v] Collapse MediaFlow Proxy section, change other settings — save should work
- [v] All existing settings should save/load correctly

### DIFF
--- a/app/Filament/Pages/Preferences.php
+++ b/app/Filament/Pages/Preferences.php
@@ -25,7 +25,10 @@ use Filament\Forms\Components\TextInput;
 use Filament\Forms\Components\Toggle;
 use Filament\Notifications\Notification;
 use Filament\Pages\SettingsPage;
+use Filament\Schemas\Components\Component;
+use Filament\Schemas\Components\EmbeddedSchema;
 use Filament\Schemas\Components\Fieldset;
+use Filament\Schemas\Components\Form;
 use Filament\Schemas\Components\Grid;
 use Filament\Schemas\Components\Group;
 use Filament\Schemas\Components\Section;
@@ -60,6 +63,23 @@ class Preferences extends SettingsPage
     public static function canAccess(): bool
     {
         return auth()->check() && auth()->user()->isAdmin();
+    }
+
+    /**
+     * Disable native browser validation on the settings form.
+     * Collapsed/hidden sections with numeric inputs trigger
+     * "An invalid form control is not focusable" in browsers
+     * because the fields are invisible but still validated.
+     * Livewire handles validation server-side anyway.
+     */
+    public function getFormContentComponent(): Component
+    {
+        return Form::make([EmbeddedSchema::make('form')])
+            ->id('form')
+            ->extraAttributes(['novalidate' => true])
+            ->footer([
+                $this->getFormActionsContentComponent(),
+            ]);
     }
 
     protected function getActions(): array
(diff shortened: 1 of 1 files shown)


gabelul:
## Tested in production ✓

Just deployed this to my live setup (m3u-editor running on Hetzner, 3 users actively streaming). The fix works — Settings page saves without issues in both Firefox and Chrome now.

**Before the fix:** Firefox would throw "An invalid form control is not focusable" every time you tried to save with the MediaFlow Proxy section collapsed or Provider Request Delay toggled off. Couldn't save *any* settings on the page without first expanding every section and filling in fields you don't even use.

**After:** Save just works. Collapsed sections stay collapsed, optional features stay optional. Livewire handles the validation server-side as it should.

Tested scenarios:
- Provider Request Delay OFF, save other settings → works
- MediaFlow Proxy section collapsed, save → works  
- Enable advanced failover logic + silence detection, save → works
- Toggle settings on/off across multiple tabs, save → works

The `novalidate` approach is the right call here — Filament/Livewire already validates everything server-side, so the native browser validation on hidden fields is just getting in the way.

sparkison:
Long running mystery solved! lol! 😂

Seriously, I've not dove into this one since it was mostly affecting Safari users and seemed intermittent.

Question, can we make this a global change? The Playlist form, for example, had this too. And there's plenty of other conditional hide/show with required that will have the same issue.

I'm wondering if we can apply this change to Livewire forms generally?

gabelul:
Ha, glad the mystery's cracked! Safari being weird about it explains the intermittent reports — each browser has its own opinion on when to enforce validation on hidden fields.

For the global fix — the cleanest approach is adding `novalidate` directly to the Filament form blade template. Since Livewire handles all validation server-side anyway, native browser validation on `<form>` elements is just noise (and actively harmful with conditional hide/show).

Two options:

**Option A — Publish and patch the Filament view (simplest):**
```bash
php artisan vendor:publish --tag=filament-schemas-views
```
Then edit `resources/views/vendor/filament-schemas/components/form.blade.php`:
```blade
<form
    {{
        $attributes
            ->merge([
                'id' => $getId(),
                'wire:submit' => $getLivewireSubmitHandler(),
                'novalidate' => true,
            ], escape: false)
            ...
```

One change, covers every form in the app. Playlist, Preferences, resources, everything.

**Option B — Upstream to Filament itself:**
This is arguably a Filament bug — any `->required()` field inside a `->collapsed()` section will trigger this. Could PR it to `filamentphp/filament` so everyone benefits. But that's a longer road.

Want me to update this PR to use Option A instead of the per-page override? It'd replace the `getFormContentComponent()` override with the published view approach, and we can drop the conditional `required()` changes too since `novalidate` handles everything.

sparkison:
Option A works for me 😉 thanks for looking into that!

sparkison:
Closing as implemented! 😉

PR #884, opened 2026-03-27
perf: Reduce Livewire table payload size

Warbs816:
## Summary

- **Gzip compression**: Added to embedded nginx config — 76%+ reduction on HTML/JSON responses. A 6MB table page compresses to well under 1MB.
- **Renderless column updates**: New `RenderlessColumnUpdates` trait prevents Livewire from re-rendering the entire table HTML when toggling a field inline (e.g. enabled toggle). Applied to Channels, VOD, and Series list pages.
- **Selective eager loading**: Relations (`epgChannel`, `playlist`) now load only the columns used by table columns, reducing model data by ~69%.

## Test plan

- [x] Load /channels with 100 rows — check browser DevTools Network tab for `Content-Encoding: gzip` header and reduced transfer size
- [x] Toggle an inline field (e.g. enabled) — verify the Livewire update response is small (no full table HTML)
- [x] Verify all table columns still render correctly (epg channel name/icon, playlist name)
- [x] Test /vods and /series pages similarly

### DIFF
--- a/app/Filament/Resources/Channels/ChannelResource.php
+++ b/app/Filament/Resources/Channels/ChannelResource.php
@@ -195,12 +195,10 @@ class ChannelResource extends Resource
                 ->toggleable(),
             ToggleColumn::make('enabled')
                 ->toggleable()
-                ->tooltip('Toggle channel status')
                 ->sortable(),
             ToggleColumn::make('can_merge')
                 ->label('Merge Enabled')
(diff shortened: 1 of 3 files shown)


Warbs816:
Removing the tooltips might be controversial! Lots of them are obvious though and it's a shed load of HTML to render on load

PR #889, opened 2026-03-28
fix: Remove scroll-to-top on table pagination page change

Warbs816:
## Summary
- Removes `setPage()` overrides in `ListChannels` and `ListVod` that dispatched a `scroll-to-top` event on every page change, causing the browser to smooth-scroll to the top
- Removes the now-unused global `scroll-to-top` JS event listener from `AppServiceProvider` and the orphaned `HtmlString` import

## Test plan
- [x] Navigate to Channels list, scroll down, change page — scroll position should be preserved
- [x] Navigate to VOD list, scroll down, change page — scroll position should be preserved
- [x] Confirm other tables (Series, Groups, Playlists, etc.) still paginate normally

### DIFF
--- a/app/Filament/Resources/Channels/Pages/ListChannels.php
+++ b/app/Filament/Resources/Channels/Pages/ListChannels.php
@@ -51,13 +51,6 @@ class ListChannels extends ListRecords
     #[Url(as: 'status')]
     public ?string $statusFilter = 'all';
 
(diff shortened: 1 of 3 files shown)


sparkison:
Good call! This was from a while back before the tables had a max height and fixed header, but it doesn't make much sense now 😉

PR #893, opened 2026-03-28
fix: dropdown menus hidden behind sticky topbar

### LINKED ISSUE #892: Unable to see Menu to Add to Custom Playlist
Running 0..10.27 dev
Want to make sure before I open a bug report... Unable to add to custom playlist...
Went to Live Channels, Channels, searched for the channels and click the Bulk Actions... 
Add to Custom Playlist is not listed... 
Did it move?

Image

Warbs816:
## Summary
- Increased `.fi-dropdown-panel` z-index from 20 to 40 so dropdown menus render above the sticky topbar (z-30)
- Added `max-height: 50vh` with `overflow-y: auto` so dropdowns become scrollable before overflowing the viewport
- Removed the previously commented-out duplicate `.fi-dropdown-panel` rule

Fixes #892

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -42,7 +42,9 @@
 }
 
 .fi-dropdown-panel {
-    z-index: 20; /* Ensure dropdowns appear above other elements */
+    z-index: 40; /* Ensure dropdowns appear above the topbar (z-30) */
+    max-height: 50vh;
+    overflow-y: auto;
 }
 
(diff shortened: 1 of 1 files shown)


sparkison:
This method doesn't work well with nested dropdowns (we've had this same "fix" implemented previously). I've create a flyout actions menu (used for Playlist) that should be able to be used for this as well.

PR #896, opened 2026-03-29
fix: auto-recount channel numbers when enabling group channels

Serph91P:
When enabling group channels (single or bulk), calculate the maximum channel number across all other enabled channels in the same playlist and recount the newly enabled group starting from max+1. This prevents duplicate channel numbers across groups in the same playlist.


sparkison:
Good call on this one, makes sense 😉

PR #901, opened 2026-03-30
fix: Use relative URLs for in-app player to prevent CORS errors

Warbs816:
## Summary
- Adds `bool $internal` parameter to `Channel::getProxyUrl()` and `Episode::getProxyUrl()` — when true, returns a relative path instead of an absolute URL via `PlaylistService::getBaseUrl()`
- All in-app player contexts (floating player, pop-out player) now use relative URLs, preventing CORS errors when the browser origin differs from `url_override` or `config('app.url')`
- External callers (playlist generation, API output) are unaffected

## Test plan
- [x] Open Playlist overview, click play on a channel — stream loads without CORS errors
- [x] Click pop-out button on floating player — pop-out player loads correctly
- [x] Play a channel from the Live Channels page — no CORS error
- [x] Play a VOD from the VOD detail page — no CORS error
- [x] Play an episode from the Series episodes list — no CORS error
- [x] Verify generated playlist URLs still use absolute URLs (external clients unaffected)

### DIFF
--- a/resources/views/filament/resources/vods/pages/view-vod.blade.php
+++ b/resources/views/filament/resources/vods/pages/view-vod.blade.php
@@ -78,7 +78,7 @@
             'content_type' => 'vod',
             'playlist_id' => $record->playlist_id,
             'title' => $title,
-            'url' => $record->getProxyUrl(username: $username, password: $password), // pass guest auth (if set)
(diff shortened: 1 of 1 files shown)


sparkison:
lol! So simple, I love it 😂

PR #905, opened 2026-03-31
fix: restore VOD and series playback in floating player

Warbs816:
## Summary
Fix floating and pop-out player playback for VOD and series content by preventing non-live streams from being forced through the MPEG-TS player path.

## Context
Series episodes and VOD items were being opened with `format: ts` and then routed into the MPEG-TS player, even when the proxied content was not actually MPEG-TS/FLV. This caused browser playback failures such as `Non MPEG-TS/FLV, Unsupported media type!` in both the in-app floating player and the pop-out player.

## Changes
- Updated the VOD view to use the model-generated floating player attributes instead of manually constructing a payload with a hardcoded format assumption.
- Adjusted frontend player selection so MPEG-TS is only used for live streams.
- Relaxed pop-out player format handling so non-HLS, non-MPEGTS formats are preserved and can fall back to native browser playback.

### Key Implementation Details
- `resources/views/filament/resources/vods/pages/view-vod.blade.php`
  - Replaced the manual `playerArgs` payload with `getFloatingPlayerAttributes($username, $password)` so VOD uses the same attribute generation path as other content.
- `resources/js/vendor/stream-viewer.js`
  - Normalizes the provided format and checks `data-content-type`.
  - Keeps HLS detection unchanged.
  - Restricts MPEG-TS initialization to `content_type === 'live'`.
  - Routes VOD and episode playback through the native player path when they are not HLS.
- `app/Http/Controllers/PlayerController.php`
  - Preserves valid alphanumeric formats passed to the pop-out player instead of coercing everything unknown back to `ts`.

## Use Cases
- Play a series episode from `/series/{id}` in the floating player.
- Play a VOD item from `/vods` in the floating player.
- Open the same content in the pop-out player without it being misclassified as MPEG-TS.

## Testing
Manual verification:
1. Open a series episode from `/series/{id}` in the floating player.
2. Open a VOD item from `/vods` in the floating player.
3. Open the same content in the pop-out player.
4. Confirm playback starts successfully and the browser no longer logs `Non MPEG-TS/FLV, Unsupported media type!` for these VOD/episode flows.

Automated checks run:
```bash
vendor/bin/pint --dirty
php artisan test
```

Notes:
- `vendor/bin/pint --dirty` passed.
- The full test suite still contains unrelated pre-existing failures and environment issues (including Redis-related failures and memory exhaustion during exception rendering), so this PR relies primarily on targeted manual playback verification for the changed behavior.

### DIFF
--- a/app/Http/Controllers/PlayerController.php
+++ b/app/Http/Controllers/PlayerController.php
@@ -25,7 +25,8 @@ class PlayerController extends Controller
         }
 
         $streamFormat = strtolower((string) $request->query('format', 'ts'));
-        if (! preg_match('/^[a-z0-9]+$/', $streamFormat)) {
+        $supportedStreamFormats = ['ts', 'mpegts', 'hls', 'm3u8', 'mp4', 'm4v', 'mkv', 'webm', 'mov'];
+        if (! in_array($streamFormat, $supportedStreamFormats, true)) {
             $streamFormat = 'ts';
         }
 
--- a/app/Http/Controllers/PlayerController.php
+++ b/app/Http/Controllers/PlayerController.php
@@ -24,8 +24,8 @@ class PlayerController extends Controller
             abort(404);
         }
 
-        $streamFormat = (string) $request->query('format', 'ts');
-        if (! in_array($streamFormat, ['ts', 'mpegts', 'hls', 'm3u8'], true)) {
+        $streamFormat = strtolower((string) $request->query('format', 'ts'));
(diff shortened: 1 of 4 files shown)


Warbs816:
Looks like this still needs some work before it's ready, regarding HLS streams.

Warbs816:
Going to scrap this and attempt again but based on the #906 changes, didn't realise I was dealing with two issues 😊

PR #908, opened 2026-03-31
fix: restore in-app player playback behavior

Warbs816:
## Summary

This PR restores correct in-app player playback behavior across floating and pop-out players.

It consolidates the earlier in-app transcoding route fixes with the follow-up playback fixes so in-app playback now consistently uses the correct route, format, and player behavior for live channels, VOD, and series episodes.

### Regression background
This fixes the in-app transcoding regression introduced by commit https://github.com/m3ue/m3u-editor/commit/dba537078148d7d728e4ecb08db8ab0c6399660f, where in-app player URL generation
   was moved onto Xtream-style proxy routes and stopped consistently following the dedicated in-app playback path.

#### In-app transcoding route usage
- restore dedicated in-app player route usage for floating/in-app playback
- ensure only in-app playback uses the in-app transcoding settings
- keep generic Xtream/proxy URLs unchanged so the in-app transcoding defaults do not leak into other playback flows

#### Remaining payload sources
- fix remaining VOD and playlist payload sources that were still emitting old Xtream-style player URLs
- align playlist live EPG playback with the same floating-player helpers already used by the corrected channel pages
- ensure VOD-related pages use the model helper instead of manually constructing stale player payloads

#### Pop-out playback behavior
- preserve native formats in the pop-out player for VOD and episode playback
- stop coercing supported native formats such as mp4 and mkv back to ts

#### Player lifecycle and playback handling
- improve player cleanup/init ordering so reinitialization does not tear down the newly bound video element
- distinguish live vs VOD/episode behavior in the in-app player configuration
- avoid applying live-only playback assumptions to VOD and episode streams

### Areas covered
- channels
- groups edit pages
- playlist live EPG playback
- vods
- vod groups
- series
- series edit pages
- floating player
- pop-out player

### Result
Floating and pop-out in-app playback now use the correct route and format behavior for live, VOD, and episodes, with in-app transcoding applied only where intended.

### DIFF
--- a/app/Http/Controllers/PlayerController.php
+++ b/app/Http/Controllers/PlayerController.php
@@ -25,7 +25,7 @@ class PlayerController extends Controller
         }
 
         $streamFormat = (string) $request->query('format', 'ts');
-        if (! in_array($streamFormat, ['ts', 'mpegts', 'hls', 'm3u8'], true)) {
+        if (! in_array($streamFormat, ['ts', 'mpegts', 'hls', 'm3u8', 'mp4', 'm4v', 'mkv', 'webm', 'mov'], true)) {
             $streamFormat = 'ts';
         }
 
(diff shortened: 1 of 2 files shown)


sparkison:
I think I got this one one 😉

PR #912, opened 2026-04-01
fix: resolve multiple Plex DVR integration issues

Serph91P:
- Fix getDvrChannels to use DVR detail endpoint (/livetv/dvrs/{id}) instead of unreliable /livetv/dvrs/{id}/channels 
- Extract channel data from Device[].ChannelMapping[] in DVR response
- Fix Lineup channel parsing: use nested Lineup[].Channel[] structure with correct field names (channelVcn, identifier)
- Prevent getDvrChannels 404 from clearing stored DVR state
- Add extractLineupChannels() helper for reliable lineup parsing
- Replace Active Sessions with DVR Sync Status verification
- Fix TVG ID warning to require 'Channel Number' instead of 'Channel ID'
- Auto-set port when server type changes (Plex=32400, Emby/Jellyfin=8096)
- Prevent duplicate DVR creation by detecting existing DVR by device key
- Reuse locally stored DVR ID if it still exists in Plex
- Update all tests to match real Plex API response structures

### DIFF
--- a/app/Filament/Resources/MediaServerIntegrations/MediaServerIntegrationResource.php
+++ b/app/Filament/Resources/MediaServerIntegrations/MediaServerIntegrationResource.php
@@ -216,6 +216,13 @@ class MediaServerIntegrationResource extends Resource
                                 ->required()
                                 ->default('emby')
                                 ->live()
+                                ->afterStateUpdated(function (Set $set, ?string $state): void {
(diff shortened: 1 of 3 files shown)


Serph91P:
@sparkison found some bugs after rolling it out yesterday on my own “prod” system, after testing it only local :wink: Those are fixed now.

PR #916, opened 2026-04-01
fix: Restore in-app player transcoding routes and playback behavior

Warbs816:
## Summary

There are a handful of related in-app player playback issues that all stem from the same regression (dba5370), so I've bundled them together here.

The main thing still outstanding is that `Channel::getFloatingPlayerAttributes()` and `Episode::getFloatingPlayerAttributes()` currently build the in-app player URL via `getProxyUrl()`, which routes playback through the generic proxy path. This means the dedicated in-app transcoding settings (stream profiles) never get applied — the player ends up bypassing transcoding entirely. This PR adds dedicated signed routes for in-app playback so that transcoding works as expected again.

### Dedicated in-app player routes (the main fix)
- **`Channel.php` / `Episode.php`**: `getFloatingPlayerAttributes()` now generates a signed URL pointing to a dedicated in-app player route instead of the generic proxy. This keeps in-app transcoding settings isolated from external proxy URLs.
- **`routes/api.php`**: Registers the two new signed routes (`m3u-proxy.channel.player`, `m3u-proxy.episode.player`).

### Playlist live EPG playback
- **`EpgApiController.php`**: The EPG grid was still manually constructing player payloads via `getProxyUrl()`. Switched to `getFloatingPlayerAttributes()` so it follows the same corrected path.

### Pop-out player VOD format support
- **`PlayerController.php`**: Adds `mp4`, `m4v`, `mkv`, `webm`, `mov` to the pop-out player's allowed format list so VOD content plays natively without being coerced to `ts`.

### VOD page payload cleanup
- **`ViewVod.php` / `view-vod.blade.php`**: Replaces manually constructed payloads with `getFloatingPlayerAttributes()` to stay consistent.

### Player lifecycle improvements (stream-viewer.js)
- Moves `cleanup()` before setting `this.player` so reinitialization doesn't tear down the newly bound video element
- Differentiates live vs VOD/episode config in HLS.js and mpegts.js (buffer sizes, latency chasing, live sync) — VOD content no longer gets treated as a live stream
- Cleaner video element teardown (`removeAttribute('src')` instead of setting `src = ''`)

### Tests
- Adds `InAppPlayerTranscodingRouteTest.php` covering live channels, VOD, and episodes — verifies that in-app playback uses the signed routes and that proxy URLs remain unaffected by transcoding profiles.

## How to verify

1. Set a stream profile with a transcoding format (e.g. `m3u8`) as the default live or VOD profile
2. Play a live channel, VOD, and episode via the floating player — should hit the `/api/m3u-proxy/channel/{id}/player` (or episode equivalent) signed route
3. Play from the playlist live EPG — should use the same corrected path
4. Check that external proxy URLs (M3U output, Xtream API) are unaffected
5. Pop out a VOD with `.mkv` / `.mp4` — should play natively

### DIFF
--- a/app/Http/Controllers/PlayerController.php
+++ b/app/Http/Controllers/PlayerController.php
@@ -25,7 +25,7 @@ class PlayerController extends Controller
         }
 
         $streamFormat = (string) $request->query('format', 'ts');
-        if (! in_array($streamFormat, ['ts', 'mpegts', 'hls', 'm3u8'], true)) {
+        if (! in_array($streamFormat, ['ts', 'mpegts', 'hls', 'm3u8', 'mp4', 'm4v', 'mkv', 'webm', 'mov'], true)) {
             $streamFormat = 'ts';
         }
 
--- a/resources/js/vendor/stream-viewer.js
+++ b/resources/js/vendor/stream-viewer.js
@@ -187,7 +187,10 @@ function streamPlayer() {
                 console.error('Video element not found:', playerId);
                 return;
             }
-                        
+
+            // Clean up any existing players before binding the new video element
+            this.cleanup();
+
             // Store reference to video element for cleanup
             this.player = video;
 
@@ -196,9 +199,6 @@ function streamPlayer() {
 
             // Reset error counters
             this.fragmentErrorCount = 0;
-
-            // Clean up any existing players
-            this.cleanup();
 
             // Initialise progress tracking from data attributes
             this._initProgress();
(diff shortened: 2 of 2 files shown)


Warbs816:
@sparkison Sorry to effectively re-open #908 but the In-App Player seems so broken for me at the moment even on v0.10.31-dev. On this branch it's flawless and respects In-App Transcoding again. I've got Chromecast functioanlity working based on these fixes, it's basically impossible without! 😊

sparkison:
I started to go down the signed routes path too, but realized it's not going to work with long running streams (especially HLS). If a user is watching an HLS stream that's over 1hr long, they will need to close and re-open the player to generate a new URL for the next hour.

Let me know what you think! I'd like to get this solved, but don't think this is quite there...

sparkison:
Realized this isn't synced up with dev either and has merge conflicts. I'll take another look, or feel free to make another PR from the latest dev without signed URLs and I'll take another look!

Warbs816:
I'll take another look tonight, I wasn't aware of the signed route issues. Cheers!

sparkison:
> I'll take another look tonight, I wasn't aware of the signed route issues. Cheers!

No worries! I think I got it 🤞

PR #917, opened 2026-04-01
Add max connections and proxy override to Playlist Auths

Warbs816:
## Summary

- Adds per-auth **max concurrent connections** (`max_connections`) — enforced when proxy is active, with support for the global "stop oldest on limit" setting
- Adds per-auth **proxy override** (`enable_proxy`) — tri-state: inherit from playlist, force on, or force off
- Updates Xtream API `user_info` response to reflect per-auth limits (`max_connections`, `active_cons`)
- Adds `username` to proxy stream metadata for per-auth stream tracking

## Files changed

- **Migration** — adds `max_connections` and `enable_proxy` columns to `playlist_auths`
- **PlaylistAuth model** — `resolveEnableProxy()`, `isAtConnectionLimit()` helpers
- **PlaylistService** — returns `PlaylistAuth` as 5th element from `authenticate()`
- **M3uProxyService** — per-auth capacity checks, `stopOldestStreamByMetadata()`, username in metadata
- **XtreamStreamController** — returns `PlaylistAuth`, `resolveEnableProxy()` helper
- **XtreamApiController** — proxy override, per-auth info in Xtream response
- **PlaylistGenerateController** — proxy override for M3U generation
- **M3uProxyApiController** — PlaylistAuth lookup and passthrough
- **PlaylistAuthResource** — form fields, table columns
- **PlaylistAuthFactory** — new field defaults

## Test plan

- [ ] Run migration
- [ ] Proxy override: test all 6 combinations (alias on/off × auth inherit/force-on/force-off)
- [ ] Select hydration: save each option, re-open edit dialog, confirm correct value shown
- [ ] Max connections: test with limit=1 and limit=0, with "stop oldest" on and off
- [ ] Xtream API: verify `max_connections` and `active_cons` in `player_api.php` response
- [ ] Edge cases: disabled auth, expired auth, unassigned auth, pooled streams

### DIFF
--- a/app/Http/Controllers/XtreamStreamController.php
+++ b/app/Http/Controllers/XtreamStreamController.php
@@ -89,7 +89,7 @@ class XtreamStreamController extends Controller
                                 }
                             }
                         } catch (ModelNotFoundException $e) {
-                            return [null, null];
+                            return [null, null, null];
                         }
                     }
                 }


Warbs816:
It's worth noting that with this change if the auth is assigned to an alias, it inherits the proxy setting from the alias, not the main playlist.

It might be worth opening a discussion around removing the proxy and auth settings from Playlist Aliases, and moving them all to Playlist Auths as it'll allow for cleaner, more unified controls, just with the caveat that if you create an Alias, you must create and assign an Auth to it to be able to access it

Again the failed test isn't mine 😊

PR #920, opened 2026-04-02
fix: Plex EPG/channel matching after recount for custom playlists

Serph91P:
Root cause: syncDvrChannels() built channel map against stale Plex lineup data because EPG refresh was triggered AFTER channel sync. Plex still had old XMLTV cached when mapping HDHR GuideNumbers.

Changes:
- EpgGenerateController: use pivot channel_number for custom playlists when id_channel_by is Number (falls back to global channel number)
- EpgGenerateController: add Cache-Control no-store headers to all 3 EPG response paths (cache HIT, compressed MISS, streaming MISS)
- PlexManagementService: reorder syncDvrChannels() to refresh EPG BEFORE channel sync loop (with 8s wait for Plex to fetch XMLTV)
- PlexManagementService: add 8s wait in addDvrDevice() before first channel sync so Plex can download XMLTV after DVR creation
- PlexManagementService: add mismatch detection logging in buildChannelMapPayload() when HDHR numbers don't match Plex lineup
- PlexManagementService: trigger POST RefreshEPGGuides in refreshGuides()
- ChannelsRelationManager, VodRelationManager, ChannelResource: clear EPG cache and dispatch SyncPlexDvrJob after channel recount
- Tests: add 3 EPG tests verifying pivot channel_number usage

### DIFF
--- a/app/Services/EpgCacheService.php
+++ b/app/Services/EpgCacheService.php
@@ -1223,6 +1223,49 @@ class EpgCacheService
     }
 
     /**
+     * Clear EPG file caches for all playlist types affected by a group channel recount.
+     *
+     * Uses the group's known playlist_id directly (skips loading channel IDs into PHP)
+     * and finds custom playlists via a JOIN on group_id.
+     */
+    public static function clearForGroup(int $groupId, int $playlistId): void
+    {
+        // Custom playlists that contain channels from this group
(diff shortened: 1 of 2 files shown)


Serph91P:
@sparkison another PR for Plex, did some more testing. Those are fixed now. Tested with my local Plex.

PR #923, opened 2026-04-02
Fix: playlist-level transcoding profiles should not apply to in-app player

Warbs816:
## Summary

Playlist-level stream profiles (set per-playlist in playlist settings) were being applied to the in-app floating/popout player when no in-app transcoding profile was configured in Preferences. This caused **broken playback in the in-app player** - the player would fail to load streams because the playlist transcoding profile was not compatible with the in-app player's proxy flow. These profiles are intended for external clients (VLC, Kodi, etc.) and should not affect in-app playback.

## Root Cause

`channelPlayer()` and `episodePlayer()` in `M3uProxyApiController` had a fallback that loaded the playlist's `streamProfile` / `vodStreamProfile` when no in-app transcoding profile was set in Preferences > In-App Player Transcoding. This caused the in-app player to unexpectedly transcode streams using the playlist's external client profile, which broke playback.

## Fix

- Removed the playlist profile fallback from both `channelPlayer()` and `episodePlayer()`. The in-app player now exclusively uses profiles configured in **Preferences > In-App Player Transcoding**. If no profile is configured, direct proxying is used (no transcoding).
- Updated the helper text on the playlist transcoding profile selects across all four playlist resources (Playlist, MergedPlaylist, CustomPlaylist, PlaylistAlias) to clarify they apply to external clients only and do not affect the in-app player.

## Files Changed

- `app/Http/Controllers/Api/M3uProxyApiController.php`
- `app/Filament/Resources/Playlists/PlaylistResource.php`
- `app/Filament/Resources/MergedPlaylists/MergedPlaylistResource.php`
- `app/Filament/Resources/CustomPlaylists/CustomPlaylistResource.php`
- `app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php`

### DIFF
--- a/app/Filament/Resources/CustomPlaylists/CustomPlaylistResource.php
+++ b/app/Filament/Resources/CustomPlaylists/CustomPlaylistResource.php
@@ -433,7 +433,7 @@ class CustomPlaylistResource extends Resource
                                 ->searchable()
                                 ->preload()
                                 ->nullable()
-                                ->helperText('Select a transcoding profile to apply to Live streams from this playlist. Leave empty for direct stream proxying.')
+                                ->helperText('Select a transcoding profile to apply to Live streams for external clients (VLC, Kodi, etc.). Does not affect the in-app player. Leave empty for direct stream proxying.')
                                 ->placeholder('Leave empty for direct stream proxying'),
                             Select::make('vod_stream_profile_id')
                                 ->label('VOD and Series Streaming Profile')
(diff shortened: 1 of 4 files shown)


Warbs816:
I changed this because when no in-app transcode was set but a playlist-level transcode was, in-app playback was broken. It also makes sense to give more control 😊

sparkison:
> I changed this because when no in-app transcode was set but a playlist-level transcode was, in-app playback was broken. It also makes sense to give more control 😊

makes sense! Thank you 🙏

PR #975, opened 2026-04-07
feat: positional index for M3U source_id and regex-based failover matching

Serph91P:
- Add 'use_positional_index' import pref for M3U playlists that includes the line number in the source_id hash, allowing duplicate entries with different URLs to be imported as separate channels for failover merging
- Add 'merge_regex' column to channels table for per-channel regex-based stream matching across playlists (e.g. /^CCTV[-]?1$/i)
- Integrate regex merge as second pass in MergeChannels job after standard stream_id grouping
- Add UI toggle in Playlist Processing settings and regex field in Channel edit form
- Fix broken Docker symlinks in test bootstrap (.env, storage/logs)
- Add tests for both features

### DIFF
--- a/app/Filament/Resources/Channels/ChannelResource.php
+++ b/app/Filament/Resources/Channels/ChannelResource.php
@@ -1290,6 +1290,16 @@ class ChannelResource extends Resource implements CopilotResource
                         ->placeholder(fn (Get $get) => $get('stream_id'))
                         ->helperText(fn (Get $get) => $get('is_custom') ? '' : 'Leave empty to use default value.')
                         ->rules(['min:1', 'max:255']),
+                    TextInput::make('merge_regex')
+                        ->label(__('Merge Regex'))
+                        ->hintIcon(
+                            'heroicon-m-question-mark-circle',
+                            tooltip: 'A regex pattern to match streams from other playlists for failover merging. Matched against the channel title and name. Example: /^CCTV[-]?1$/i'
+                        )
+                        ->columnSpan(1)
+                        ->placeholder('/^CCTV[-]?1$/i')
+                        ->helperText(__('Regex pattern for cross-playlist failover matching.'))
+                        ->rules(['max:500']),
                     TextInput::make('station_id')
                         ->label(__('Station ID'))
                         ->hint(__('tvc-guide-stationid'))
--- a/app/Filament/Resources/Playlists/PlaylistResource.php
+++ b/app/Filament/Resources/Playlists/PlaylistResource.php
@@ -1771,6 +1771,18 @@ class PlaylistResource extends Resource implements CopilotResource
                         ->helperText(__('When enabled, groups will be included based on regex pattern match instead of prefix.'))
                         ->hidden(fn (Get $get): bool => ! $get('import_prefs.preprocess') || ! $get('status')),
 
+                    Toggle::make('import_prefs.use_positional_index')
+                        ->label(__('Use positional index for channel identity'))
+                        ->columnSpan(2)
+                        ->inline(true)
+                        ->default(false)
+                        ->hintIcon(
+                            'heroicon-m-question-mark-circle',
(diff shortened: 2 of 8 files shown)


sparkison:
I'm not quite sure on this one yet. And thinking we could simply add _just_ the index option to the `source_id` generation for uniqueness, I think that's really all that's needed to fix this one.

Then the merge should work as expected since it matches on `stream_id`, and the `stream_id` will be set to the M3U `tvg-id`, which are all the same in the example issue:

```
#EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
https://example.com/channel/cctv1?streamid=aaaaa&livekey=123
#EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
https://example.com/channel/cctv1?streamid=bbbbb&livekey=123
#EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
https://example.com/channel/cctv1?streamid=ccccc&livekey=123
```

...so basically, thinking we can drop all the `merge_regex` stuff, it doesn't seem like it's needed. At least not to me, unless I'm missing something 🤷

Serph91P:
> I'm not quite sure on this one yet. And thinking we could simply add _just_ the index option to the `source_id` generation for uniqueness, I think that's really all that's needed to fix this one.
> 
> Then the merge should work as expected since it matches on `stream_id`, and the `stream_id` will be set to the M3U `tvg-id`, which are all the same in the example issue:
> 
> ```
> #EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
> https://example.com/channel/cctv1?streamid=aaaaa&livekey=123
> #EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
> https://example.com/channel/cctv1?streamid=bbbbb&livekey=123
> #EXTINF:-1 tvg-id="CCTV1" tvg-name="CCTV1" group-title="CCTV",CCTV-1
> https://example.com/channel/cctv1?streamid=ccccc&livekey=123
> ```
> 
> ...so basically, thinking we can drop all the `merge_regex` stuff, it doesn't seem like it's needed. At least not to me, unless I'm missing something 🤷

You're right that the positional index alone solves the original issue - within a single playlist, all three CCTV1 entries share the same tvg-id, so once they're imported as separate channels (via the index), the existing stream_id-based merge groups them correctly.

However, the merge_regex addresses a second, separate problem that @KazeLiu raised further down in #974 - cross-playlist name normalization:

When subscribing to multiple sources, the same channel (e.g., CCTV-1) often has wildly different names: CCTV1, CCTV-1, cctv1综合, cctv1综合频道, cctv1綜合頻道

In this scenario, different providers use completely different tvg-id values for the same channel (e.g. Provider A uses CCTV1, Provider B uses cctv_general, Provider C uses cn-cctv1). Since the stream_id values are all different, the standard merge will never group them together. The regex matcher is the only way to bridge that gap - the user defines a pattern like /^CCTV[-]?1(综合(频道)?)?$/i on the master channel, and the second pass matches candidates by title/name across all playlists.

So in short:

Positional index - fixes the original bug (same-playlist duplicates with identical metadata)
merge_regex - addresses the follow-up request (cross-playlist matching where providers use different names/IDs for the same channel)
They solve different problems. If you want, I can split the PR into two - one for just the positional index fix, and a separate one for the regex merge feature.

sparkison:
ahh! Sorry, totally missed the follow up. Thanks for clearing it up. I'll take another look...

sparkison:
ok, working on something slightly different! 

Works in a similar way, but without the positional index. It's a look-back look-ahead dupe check within the channel import bunch, and if there's a duplicate found, it's tagged with a unique key. This keeps the existing imports working as they do now (the positional index isn't forced, and the provider changing order won't affect it).

I also moved the regex up to the playlist level to make it easier to manage. It should still achieve the same end result, just doing it a little different 😉 

merge-patterns

PR #982, opened 2026-04-07
fix: Add rate limiting to watch progress endpoints

Warbs816:
## Summary
- Wraps the watch progress GET/POST routes in `throttle:60,1` middleware
- Matches the same rate limiting already applied to the EPG API routes
- Controller already handles auth internally, so this just adds the missing abuse protection

## Test plan
- [x] Play a VOD/episode — watch progress saves and resumes correctly
- [x] Play a live channel — tune-in is recorded
- [x] Automated tests verify middleware is applied to both routes


Warbs816:
Failed tests are pre-existing, not mine 😊

PR #985, opened 2026-04-07
chore: Remove debug console.log calls from player code

Warbs816:
## Summary
- Removes ~30 `console.log` calls from `stream-viewer.js` and `multi-stream-manager.js`
- Keeps `console.warn`/`console.error` in catch blocks for genuine error conditions
- Removes two HLS event handlers (`FRAG_LOAD_ERROR`, `LEVEL_LOADED`) that existed purely for debug logging

## Test plan
- [x] Open a stream (HLS, MPEG-TS, native) — plays as before, console is clean
- [x] Trigger a stream error (bad URL) — error overlay appears, `console.warn`/`console.error` still fires
- [x] Close a player — cleanup completes silently

### DIFF
--- a/resources/js/vendor/multi-stream-manager.js
+++ b/resources/js/vendor/multi-stream-manager.js
@@ -39,7 +39,6 @@ function multiStreamManager() {
                 if (Array.isArray(detail)) {
                     detail = detail[0];
                 }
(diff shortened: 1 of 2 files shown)


Warbs816:
IMO doesn't make sense to have these in production

PR #986, opened 2026-04-07
feat: Add configurable max concurrent floating players limit

Warbs816:
## Summary
- Adds a "Max Concurrent Floating Players" setting to Preferences > Proxy > In-App Player Transcoding section
- Default limit is 4, set to 0 for unlimited
- When the limit is reached, a small tooltip appears near the clicked button for 2 seconds then fades out
- Setting is stored via Spatie Laravel Settings with a migration

## Test plan
- [x] Run `php artisan migrate` to apply the new setting
- [x] Preferences > Proxy > In-App Player Transcoding — new field visible, defaults to 4
- [x] Open 4 floating players, try a 5th — tooltip appears near the button, player doesn't open
- [x] Close one, open another — works again
- [x] Set to 0 — unlimited, can open as many as you want
- [x] Set to 1 — only one player at a time
- [x] Opening a duplicate channel still brings existing player to front (not affected by limit)


Warbs816:
Wasn't sure where the best place for the setting is.. feel free to move it 😊

PR #987, opened 2026-04-08
feat: Plugin update checker with GitHub release notifications

Serph91P:
- Add PluginUpdateChecker service that queries GitHub Releases API
- Add plugins:check-updates artisan command with --plugin and --force options
- Add scheduled update check (configurable frequency via PLUGIN_UPDATE_CHECK_FREQUENCY)
- Add migration with update-check columns on extension_plugins table
- Add repository field to PluginManifest with URL normalization
- Add repository format validation in PluginValidator
- Add hasUpdateAvailable() and isFromGithub() to Plugin model
- Add stageUpdateFromLatestRelease() to PluginManager
- Add update badge on plugin version column in PluginResource table
- Add check_for_updates and stage_update actions on EditPlugin page
- Add Check for Updates header action and Updates Available card on PluginsDashboard
- Add SHA-256 resolution from companion files and release body parsing
- Add config/plugins.php update_check section (enabled, frequency, github_token)
- Add 21 tests covering API calls, version comparison, SHA-256, command, manifest

### DIFF
--- a/app/Console/Commands/CheckPluginUpdates.php
+++ b/app/Console/Commands/CheckPluginUpdates.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\Plugin;
+use App\Plugins\PluginUpdateChecker;
+use Illuminate\Console\Command;
+
+class CheckPluginUpdates extends Command
+{
+    protected $signature = 'plugins:check-updates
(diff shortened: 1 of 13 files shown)


Serph91P:
@sparkison I am lazy and that's why at least a notification for an update would be good. Then I thought why not just update it as well?

PR #992, opened 2026-04-08
Fix mixed content errors for package assets behind reverse proxy

Warbs816:
## Summary
- Widened HTTPS detection gate from checking only `X-Forwarded-Proto` to running for all HTTP requests (covers all 6 proxy header variants + `APP_URL` fallback)
- Moved scheme detection from `boot()` to `register()` to fix a boot order issue where package service providers (e.g. filament-copilot) call `asset()` during their `boot()`, baking in `http://` before our `URL::forceScheme('https')` ran

## Test plan
- [x] Load page over HTTPS — confirm no mixed content warnings in browser console
- [x] Verify filament-copilot CSS/JS load over HTTPS in the network tab
- [x] Confirm plain HTTP access still works if applicable

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -94,6 +94,14 @@ class AppServiceProvider extends ServiceProvider
 
         $this->app->singleton(GitInfoService::class);
 
+        // Detect HTTPS early in register() so that package service providers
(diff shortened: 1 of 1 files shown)


sparkison:
Implemented! 😉

PR #994, opened 2026-04-08
feat: Pop-in from pop-out player back to floating player

Warbs816:
## Summary
- Adds a "Pop In" button to the pop-out player header that sends the stream back to the source tab as a floating player
- Uses BroadcastChannel with source tab targeting so the stream returns to the exact tab it came from
- Heartbeat ping/pong greys out the button with cursor-not-allowed if the source tab has been closed
- VOD/episodes carry their current playback position so the floating player resumes from the same point
- Skips proxy stop notification on pop-in transfer so the stream stays alive during handoff
- Button is hidden entirely if the pop-out was opened directly (not from a floating player)

## Test plan
- [x] Open floating player, pop out, click "Pop In" — stream returns to the same tab as a floating player
- [x] With multiple m3u-editor tabs open, pop-in only goes to the source tab
- [x] Close the source tab — pop-in button greys out within a few seconds
- [x] VOD: pop-in resumes from the same playback position
- [x] Live: pop-in works without interruption
- [x] Proxy logs show no stream stop/restart during the transfer
- [x] Open a pop-out directly via URL — no pop-in button shown

### DIFF
--- a/resources/js/vendor/multi-stream-manager.js
+++ b/resources/js/vendor/multi-stream-manager.js
@@ -68,14 +68,27 @@ function multiStreamManager() {
             // Reposition players when viewport shrinks
             window.addEventListener('resize', () => this.constrainAllToViewport(), { signal });
 
-            // Listen for pop-in requests from pop-out windows
+            // Listen for pop-in requests from pop-out windows (two-phase handshake)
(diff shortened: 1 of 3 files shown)


sparkison:
I'm going to skip this one for now unless it comes up as an issue/feature request, then we can revisit.

PR #1000, opened 2026-04-09
fix: Handle null backdrop paths in Xtream API series/VOD responses

### LINKED ISSUE #999: 500 error for get_series on xtream api
## Describe the bug

Getting 500 error on xtream for actions get_series

## To Reproduce

Not sure how to reproduce only one playlist seem to be having the issue.

## Expected behavior

get_series to return series

## Actual behavior
series blank on player

Debug log got this entry 
App\Http\Controllers\LogoProxyController::generateProxyUrl(): Argument #1 ($originalUrl) must be of type string, null given

Warbs816:
Fixes #999

## Summary

- Accept nullable `$originalUrl` in `LogoProxyController::generateProxyUrl()` so null values reach the existing empty/invalid URL check (returns placeholder) instead of throwing a TypeError
- Filter null entries from `backdrop_path` arrays with `array_filter()` before processing in `get_series`, `get_series_info`, and `get_vod_info` actions
- Add test covering `get_series` with logo proxy enabled and null backdrop path entries

## Root cause

Series metadata from Xtream providers can contain null elements in the `backdrop_path` JSON array. When logo proxy is enabled, `array_map` passes these nulls to `generateProxyUrl(string $originalUrl)`, which throws a TypeError before the method's own empty-check logic can handle it.

### DIFF
--- a/app/Http/Controllers/LogoProxyController.php
+++ b/app/Http/Controllers/LogoProxyController.php
@@ -70,7 +70,7 @@ class LogoProxyController extends Controller
     /**
      * Generate a proxy URL for a given logo URL
      */
-    public static function generateProxyUrl(string $originalUrl, $internal = false): string
+    public static function generateProxyUrl(?string $originalUrl, $internal = false): string
     {
         // Get the config values (takes priority over settings values)
         $proxyUrlOverride = config('proxy.url_override');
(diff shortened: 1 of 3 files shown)


Warbs816:
#1000 :)

PR #1007, opened 2026-04-10
feat: Add Simplified Chinese translation for app

sparkison:
Includes initial Google Translate values.

@KazeLiu, please review and make adjustments as needed 😉


KazeLiu:
@sparkison sparkison Thanks for the response! My plan is to polish and refine the existing Google-translated content to ensure better quality. I will follow the same workflow as before:

Refine the README file.

Update zh_CN.json based on en.json.

Modify the files in the zh_CN folder.

If this approach works for you, I'll get started right away.

PR #1014, opened 2026-04-11
feat: Add regex extract mode and append suffix to EPG mapping

### LINKED ISSUE #423: EPG Mapping Regex syntax + append text (post regex)
When dealing with huge lists of channels (such as local network channels in a country), auto mapping to public EPGs doesn't work well (eg. EPGShare01 US Locals).  

Applying Regex to strip out everything but 4 capital letters in a row (and alone, not part of a wider word) doesn't seem to help much; as often it will be the 4 letters and then a hypen with DT/LD/CS after it.  

It would be great to allow for the regex to remove patterns from the channel information (to get down to the 4 letter station ID) then be able to add text to the end.  

So:

 CBS 123 (OHIU) Local is parsed with the regex "(?<![A-Z])([A-Z]{4})(?![A-Z])" returns only "OHIU".  Now I need to take OHIU and automatically add -DT to the end; hoping that would now result a match in the EPGSHARE01 US Locals mapping.

I've done a series of regex methods to get to the 4 letter code; but that still won't match for whatever reason - only thing I can think of is how the EPG has -DT/LD/CS at the end of each channel (at least when I manually map this in the tool, that's how it comes up - so it does find it with the manual 4 letter search but it doesn't auto map).

**Additional context**

Warbs816:
## Summary
- Adds a **regex extract mode** toggle to EPG mapping advanced settings — when enabled, the first capture group from the regex pattern is used as the channel value instead of removing the match (e.g. `(?<![A-Z])([A-Z]{4})(?![A-Z])` on `CBS 123 (OHIU) Local` extracts `OHIU`)
- Adds an **append suffix** text field that appends text to cleaned channel values before EPG lookup (e.g. appending `-DT` to get `OHIU-DT`)
- No migrations needed — settings stored in existing JSON `settings` column on `epg_maps`

Closes #423

## Test plan
- [x] Create an EPG map with regex enabled + extract mode on, using a capture group pattern
- [x] Verify the extracted value is used for matching instead of removing the match
- [x] Add an append suffix and verify it's appended before EPG lookup
- [x] Verify default regex mode (extract off) still removes matched text as before
- [x] Verify append suffix works independently without regex


sparkison:
Closing this one down! I've opted for a slightly different path by including a local station lookup automatically in the EPG mapper steps. No config or settings changes required 😉

PR #1015, opened 2026-04-11
feat: Add optional series metadata to M3U output

Warbs816:
## Summary
- Adds a new `include_series_metadata_in_m3u` toggle to playlists, visible when "Include series in M3U output" is enabled
- When enabled, appends `tvg-season` and `tvg-episode` attributes to series episode EXTINF lines in M3U output
- Helps external tools (e.g. Emby) correctly identify episodes from the M3U file

Closes #399

## Test plan
- [x] Enable "Include series in M3U output" on an Xtream playlist — verify the new metadata toggle appears
- [x] Enable the metadata toggle, save, and fetch the M3U output — verify `tvg-season` and `tvg-episode` attributes appear on episode lines
- [x] Disable the metadata toggle — verify those attributes are absent from the output
- [x] Verify regular channel and VOD output is unaffected

PR #1016, opened 2026-04-11
fix: Fall back to series TMDB ID for episodes in M3U output

Warbs816:
## Summary
- Episode `tmdb_id` is rarely populated — it requires explicit TMDB metadata fetching
- The TMDB ID is typically available on the parent series, either in the `tmdb_id` column or `metadata.tmdb` from the provider import
- Extends the fallback chain: episode `tmdb_id` -> episode `info.tmdb_id` -> series `tmdb_id` -> series `metadata.tmdb_id` -> series `metadata.tmdb`

## Test plan
- [x] Enable "Include series in M3U output" on a playlist with series that have provider metadata
- [x] Fetch the M3U output and verify `tmdb-id` attributes now appear on episode EXTINF lines
- [x] Verify VOD/channel `tmdb-id` output is unaffected

### DIFF
--- a/app/Http/Controllers/PlaylistGenerateController.php
+++ b/app/Http/Controllers/PlaylistGenerateController.php
@@ -332,7 +332,7 @@ class PlaylistGenerateController extends Controller
                             }
 
                             $extInf = "#EXTINF:$runtime";
-                            $episodeTmdbId = $episode->tmdb_id ?: ($episode->info['tmdb_id'] ?? null);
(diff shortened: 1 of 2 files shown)


Warbs816:
Noticed this when working on #1015!

sparkison:
Good catch! I used a slightly different approach, using the `->getMovieDbIds()` https://github.com/m3ue/m3u-editor/commit/35444cc68806fbca1bb692ba53517704a97e2640

PR #1019, opened 2026-04-13
docs: add README-CN.md and link it in the original README

KazeLiu:
I'd like to submit the Chinese README first, keeping it separate from the zh_CN.json translation in two different PRs. Since the content is quite large, I wanted to start with this PR to get familiar with the workflow (this is my first time contributing on GitHub 😄). I’m a bit worried about making mistakes and want to ensure everything is correct.

Please let me know if any further changes are needed.

### DIFF
--- a/README-ZH.md
+++ b/README-ZH.md
@@ -1,0 +1,100 @@
+<div align="center">
+
+  <img src="./public/favicon.png" width="96">
+
+  <h3>m3u editor</h3>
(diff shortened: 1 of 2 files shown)


sparkison:
Looks good! I made one small tweak to hide the links for the current page (e.g. English link disabled when looking at English README and same for Chinese README).

PR #1022, opened 2026-04-13
fix: use ProxyFacade for Plex DVR URL generation and store hdhr_base_url per tuner

Serph91P:
- Store hdhr_base_url in plex_dvr_tuners JSON per tuner entry so sync uses the same network path that was verified during DVR registration
- Auto-fill HDHR/EPG URLs in DVR setup modal using ProxyFacade::getBaseUrl() which respects url_override and is request-aware via url('')
- syncDvrChannelsForTuner() and verifyDvrChannels() prefer stored URL, fall back to resolvePlaylistUrls() for legacy tuners
- Remove getAvailableBaseUrls() IP detection — not needed since ProxyFacade already handles override/request-aware resolution
- Add 3 tests covering stored URL usage and backward compatibility


Serph91P:
@sparkison a little bit of qol

PR #1025, opened 2026-04-13
fix: Session-guarded player stream stop with stall recovery

Warbs816:
## Summary
- Prevents the floating player's stop action from killing proxy streams that other sessions are still using (other tabs, popout players, external clients)
- Uses Redis SET operations (SADD/SREM/SCARD) to track active player sessions per stream — the proxy stream is only stopped when the last in-app session unregisters
- Adds automatic stall recovery for mpegts streams — when a shared stream ends and the player stalls, it detects the stall after 5 seconds and reconnects with a fresh connection (capped at 3 attempts)

## Test plan
- [x] Open same channel in two floating players — closing one should not kill the other's stream
- [x] Open a floating player then pop it out — closing the floating player should not kill the popout stream
- [x] Close the last player for a channel — proxy stream should be stopped as before
- [x] Legacy clients without player_session should still trigger proxy stop (backwards compatible)
- [x] Stall recovery: if an mpegts stream stalls mid-playback, it should auto-reconnect within ~5-10 seconds

### DIFF
--- a/app/Http/Controllers/Api/M3uProxyApiController.php
+++ b/app/Http/Controllers/Api/M3uProxyApiController.php
@@ -18,6 +18,7 @@ use Exception;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Redis;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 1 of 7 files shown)


sparkison:
We've gotten away from using Redis keys for this type of tracking is it tends to drift and is not easy to keep in sync. I'm working on an update to the proxy to allow stopping with a force flag (default true so it works as it does now). This way we can stop only if no clients are still connected by passing force = false to the proxy when stopping a stream. This should be much more consistent and leaves the proxy as the source of truth without needing to handle connection tracking in the editor.

PR #1028, opened 2026-04-14
fix: VOD TMDB filters check dedicated column + job query optimization

Serph91P:
- Fix has_tmdb_id and missing_tmdb_id filters in VodResource to also check the dedicated channels.tmdb_id and channels.imdb_id columns (previously only checked info/movie_data JSON fields, causing channels with IDs only in the dedicated column to appear as 'missing')
- Optimize buildVodQuery() and buildSeriesQuery() in FetchTmdbIds: when overwriteExisting=false, exclude fully-processed items at DB level (items with last_metadata_fetch set + tmdb_id set + complete metadata) so they are never loaded into memory, reducing job overhead on large playlists without skipping valid metadata-refresh candidates
- Add tests verifying the query optimization excludes done items and includes items that still need metadata


Serph91P:
@sparkison I hope this is ok, tried to optimize this some more.

PR #1030, opened 2026-04-14
fix: correct EPG URL path in Plex tuner registration form\n\nThe auto…

Serph91P:
…-filled EPG URL was using /epg/{uuid} which doesn't match\nany route. The correct route is /{uuid}/epg.xml. This caused Plex\nto fail resolving the lineup and report the tuner as unreachable."


Serph91P:
@sparkison another bug I found

PR #1038, opened 2026-04-17
perf: Batch pre-fetch models in Stream Monitor to eliminate N+1 queries

Warbs816:
## Summary
- The Stream Monitor page was calling `Channel::find()`, `Episode::find()`, `StreamProfile::find()`, `PlaylistProfile::find()`, and `Network::where('uuid', ...)->first()` individually inside the stream/broadcast render loops. With auto-refresh running every 5s, each open tab was firing ~5N+3 queries per tick.
- Refactored each lookup to follow the existing `whereIn()->get()->keyBy()` pattern already used for `Playlist` and `PlaylistAlias` a few lines above, dropping a refresh with 20 streams + 5 broadcasts from ~50 queries to ~7.
- No UI or behaviour changes — lookup misses still resolve to `null` via `$map[$id] ?? null`, matching the prior `find()` semantics.

## Test plan
- [x] Open `/admin/proxy/stream-monitor` with active streams and confirm channel/episode titles + logos still render on each stream card
- [x] Verify the transcoding badge still shows format + backend (FFmpeg/Streamlink/yt-dlp) on transcoded streams
- [x] Verify the provider profile badge still appears where applicable
- [x] Confirm the broadcasts section shows the real network name and HLS URL (not a raw UUID)
- [x] Trigger Failover and Remove Stream actions still behave as before
- [ ] With the proxy stopped, the connection-error banner still appears cleanly


sparkison (commented):
Looks good, thank you!

PR #1039, opened 2026-04-17
Verify stream ownership before Stream Monitor failover/stop actions

Warbs816:
## Summary
- Adds an ownership guard to `triggerFailover` and `stopStream` on the Stream Monitor page so a user can only act on streams/broadcasts they own.
- Stream ownership is checked against the filtered `fetchActiveStreams()` set (which already scopes by `getAllPlaylistUuids()`); broadcast ownership is checked against `Network::user_id`.
- Blocked calls emit a Filament notification and a `Log::warning('Unauthorized stream-monitor action blocked', ...)` entry with `user_id` and `stream_id`.

## Why
Both Livewire actions forwarded the caller-supplied `$streamId` straight to the proxy service with no check that the caller owned the stream. A user with `use_proxy` could call either action against any stream ID (including broadcast UUIDs) by issuing the Livewire call from devtools, affecting another user's stream.

The guard lives at the Livewire boundary so the service stays reusable. It fails closed — if ownership can't be verified, the action is blocked.

## Verified in browser (live)
Against the deployed container running this branch:

| Path | Result |
| --- | --- |
| `triggerFailover('fake-stream-id-...')` | Blocked, "Not authorized to manage this stream." notif, warning logged |
| `stopStream('broadcast:00000000-...')` | Blocked, notif + warning logged |
| `triggerFailover('<own stream>')` | Allowed, success notif, status badge flipped `Failover Ready` → `Failover Active` |

## Test plan
- [x] Sanity-check the page still loads and lists only streams/broadcasts you own
- [x] Click `Trigger Failover` / `Remove Stream` on your own stream — happy path works
- [x] Open devtools on a second account and try `Livewire.all().find(c => c.name.endsWith('M3uProxyStreamMonitor')).$wire.call('triggerFailover', '<someone-else-stream-id>')` — should show the "Not authorized" notification and leave a warning in `storage/logs/laravel-*.log`

### DIFF
--- a/app/Filament/Pages/M3uProxyStreamMonitor.php
+++ b/app/Filament/Pages/M3uProxyStreamMonitor.php
@@ -210,7 +210,7 @@ class M3uProxyStreamMonitor extends Page
     private function authorizeStreamAction(string $streamId): bool
     {
         if (str_starts_with($streamId, 'broadcast:')) {
-            $networkUuid = substr($streamId, 10);
+            $networkUuid = substr($streamId, strlen('broadcast:'));
             $owned = Network::where('uuid', $networkUuid)
                 ->where('user_id', auth()->id())
                 ->exists();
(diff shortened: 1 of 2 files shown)


sparkison (commented):
I might add an admin check 🤔 just so admins can stop any stream. Looks good otherwise, thanks again! 🙏

Warbs816:
Happy to pick that up!

sparkison:
> Happy to pick that up!

All good! Confirmed we're scoping streams to the user, so we'll keep it like it is.

PR #1046, opened 2026-04-19
Stream Monitor: spinner in 'Last updated' badge while refreshing

Warbs816:
## Summary

- While any `refreshData` call is in flight (auto-refresh tick, manual **Refresh**, failover/stop actions that re-fetch), the trailing timestamp in the *Last updated* badge swaps to a small spinner and back.
- The "Last updated:" label stays static and the badge stays on a single line — only the timestamp glyph is replaced by the spinner.
- Uses Livewire's built-in `.delay` (~200ms) threshold via `wire:loading.delay` / `wire:loading.delay.remove`, so fast responses never flash the spinner.
- Pure Blade/Livewire markup — no component changes, no new JS.

## Test plan

- [x] On `/m3u-proxy-stream-monitor`, with auto-refresh on and the proxy slow, the badge briefly swaps the timestamp for a spinner every 5s, then returns to the time
- [x] With a fast proxy, no visible spinner flicker on polls
- [x] Manual **Refresh** header action triggers the same swap
- [x] Failover / stop actions (which call `refreshData` afterwards) trigger the swap
- [x] Badge stays on a single line in both idle and loading states


sparkison:
Gonna skip this one, it kept flashing for me, and think I prefer the simple update instead without load indicator.

PR #1060, opened 2026-04-21
✨ feat: DVR recording engine

Grimothy:
## Summary

Full DVR recording engine built on top of Laravel + FFmpeg, integrated into the existing playlist/VOD/EPG infrastructure.

### Core Recording Engine
- **`DvrRecorderService`** — spawns FFmpeg processes from queued jobs, tracks recording state (Scheduled → Recording → PostProcessing → Completed/Failed)
- **`DvrSchedulerService`** — ticks every minute, matches EPG programme schedules against `DvrRecordingRule` entries and creates `DvrRecording` rows
- **`DvrPostProcessorService`** — runs after recording stops: ffmpeg concat demuxer for segment stitching, then dispatches metadata enrichment and VOD integration
- **`DvrRetentionService`** — cleans up old recordings based on configurable retention policy

### Metadata Enrichment
- **`DvrMetadataEnricherService`** — enriches recordings with TMDB (TV + movie search) and TVMaze (free fallback); normalises Unicode-decorated titles before searching
- Per-DvrSetting TMDB API key, falls back to global GeneralSettings key, then `config/dvr.php`

### VOD Integration
- **`DvrVodIntegrationService`** — converts completed recordings into VOD library entries:
  - TMDB `type=movie` (or no metadata + no season) → `Channel` with `is_vod=true`
  - TMDB `type=tv` (or season set) → `Series / Season / Episode`
  - TVMaze used as fallback for series name, cover, and plot when TMDB is absent
  - Recordings without any metadata get the recording date appended to their title (e.g. `"CNN News — Apr 21, 2026"`) so repeated recordings are individually identifiable
  - All DVR VOD entries placed in a "DVR Recordings" group/category
  - `dvr_recording_id` FK on `channels` and `episodes` for idempotent re-integration

### Filament UI
- **DVR Recording Rules** resource — create/edit/delete schedule rules (title match, channel, rule type, once/recurring)
- **DVR Recordings** resource — list, view, retry post-processing, cancel, delete
- **DVR Settings** tab on Playlist edit page — enable DVR, retention days, TMDB key, proxy toggle, metadata enrichment toggle
- Action menus use `RecordActionsPosition::BeforeCells` consistent with all other resources

### EPG Integration
- `EpgProgramme` model + `EpgCacheService` — caches programme schedules from EPG feeds for scheduler matching
- `EpgViewer` Livewire component with category/group tabs

### Jobs & Queues
- `StartDvrRecording`, `StopDvrRecording`, `PostProcessDvrRecording`, `EnrichDvrMetadata`, `IntegrateDvrRecordingToVod`, `DvrSchedulerTick`, `DvrRetentionCleanup`
- Dedicated Horizon queues: `dvr`, `dvr-post`, `dvr-meta`

### Other
- `use_proxy` toggle on `DvrSetting` — controls whether the DVR recorder uses the playlist proxy URL
- `EnrichDvrRecordings` Artisan command for backfilling metadata on existing recordings
- Stream endpoint: `GET /dvr/recordings/{uuid}/stream`

## Tests

65 feature tests covering: scheduler, recorder service, VOD integration, retention, and Filament table filters. All passing (1 pre-existing Redis failure in `ChannelObserverTest` unrelated to this PR).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,6 +25,7 @@ storage/logs/*
 
 # Data directories
 data/
+local/
 db-dumps/
 database/*.sqlite*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,10 @@ COPY package.json package-lock.json ./
 
 # Install all dependencies including dev deps (Vite is needed for build)
 # Note: NODE_ENV must NOT be set to production here, or npm ci will skip devDependencies
-RUN npm ci --silent
+# npm ci uses the lock file, which may contain only the glibc optional dep
+# (@rollup/rollup-linux-x64-gnu). Running npm rebuild afterwards forces npm to
+# resolve and install the correct musl variant for this Alpine environment.
+RUN npm ci --silent && npm rebuild
 
 # Copy only files needed for the build
 COPY vite.config.js postcss.config.js ./
--- a/app/Console/Commands/EnrichDvrRecordings.php
+++ b/app/Console/Commands/EnrichDvrRecordings.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Enums\DvrRecordingStatus;
+use App\Jobs\EnrichDvrMetadata;
+use App\Jobs\IntegrateDvrRecordingToVod;
+use App\Models\DvrRecording;
+use Illuminate\Console\Command;
+
+class EnrichDvrRecordings extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'dvr:enrich
+                            {--id= : Enrich a specific recording by ID}
+                            {--all : Re-enrich all completed recordings, even those already enriched}
+                            {--integrate-only : Skip metadata enrichment, only re-run VOD integration}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Enrich completed DVR recordings with metadata and integrate them into the VOD library';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): int
+    {
+        $query = DvrRecording::where('status', DvrRecordingStatus::Completed)
+            ->whereNotNull('file_path');
+
+        if ($this->option('id')) {
+            $query->where('id', (int) $this->option('id'));
+        } elseif (! $this->option('all')) {
+            // Default: only recordings missing metadata or missing a VOD entry
+            $query->where(function ($q) {
+                $q->whereNull('metadata')
+                    ->orWhereJsonLength('metadata', 0)
+                    ->orWhereDoesntHave('vodChannel')
+                    ->orWhereDoesntHave('vodEpisode');
+            });
+        }
+
+        $recordings = $query->get();
+
+        if ($recordings->isEmpty()) {
+            $this->info('No recordings to enrich.');
+
+            return self::SUCCESS;
+        }
+
+        $this->info("Dispatching enrichment for {$recordings->count()} recording(s)...");
+
+        $integrateOnly = $this->option('integrate-only');
+
+        foreach ($recordings as $recording) {
+            $label = "[{$recording->id}] {$recording->title}";
+
+            if ($integrateOnly || ! empty($recording->metadata)) {
+                // Metadata already present (or skipped) — go straight to VOD integration
+                IntegrateDvrRecordingToVod::dispatch($recording->id)->onQueue('dvr-post');
+                $this->line("  → Queued VOD integration for {$label}");
+            } else {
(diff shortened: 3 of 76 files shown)


sparkison:
## Review Notes — DVR Proxy Architecture (follow-up fixes)

This comment covers the changes made on top of the initial DVR engine commit to address issues found during testing.

---

### Architecture changes

**DVR recording now fully proxy-backed**
The original PR managed FFmpeg processes directly from PHP using `proc_open` and PID tracking. All recording is now delegated to the m3u-proxy via its `/broadcast/{uuid}/start` endpoint with `dvr_mode: true`. PHP no longer spawns or signals any FFmpeg processes.

- `DvrRecorderService` — replaced `proc_open`/PID management with `M3uProxyService::startDvrBroadcast()` and `stopDvrBroadcast()`
- `M3uProxyService` — added `startDvrBroadcast()`, `stopDvrBroadcast()`, `getDvrBroadcastLiveUrl()`, `getDvrCallbackUrl()`
- `DvrCallbackController` (new) — receives `programme_ended` / `broadcast_failed` webhooks from the proxy; transitions recording status and dispatches `PostProcessDvrRecording`
- `dvr_recordings` table — added `proxy_network_id` column; `pid` column kept but no longer written to
- **Callback URL uses the internal resolver address** (`failoverResolverUrl`), not the public URL, so it reaches the editor from inside Docker. `startDvrBroadcast()` throws if this isn't configured.

**DVR settings simplified**
Removed fields that no longer apply now that the proxy manages FFmpeg:
- Removed "Use Proxy for Recordings" toggle (always true)
- Removed "FFmpeg Path" field
- Removed all transcoding options (re-encode, codec, preset, bitrate)
- Added single **Output Format** selector (`ts` / `mp4` / `mkv`) — all three use stream copy, no re-encoding. MPEG-TS is a binary segment concat (no FFmpeg); MP4/MKV use `ffmpeg -c copy`
- Replaced per-playlist TMDB API key field with a global-key status indicator linking to Settings

**Post-processing (proxy → editor)**
- `DvrPostProcessorService` — handles MPEG-TS (binary concat, no FFmpeg) and MP4/MKV (FFmpeg `-c copy`)
- Correctly resolves the HLS manifest as an absolute path (proxy sends `hls_dir` as an absolute filesystem path; previously the code was prepending Laravel's storage root to it)
- Manifest filename corrected from `stream.m3u8` → `live.m3u8` (matching the broadcast manager's output)

**Stream Monitor**
- Active DVR broadcasts now appear in the Stream Monitor with a red **DVR** badge
- Fixed `fetchBroadcasts()` which was filtering out DVR broadcasts (they use the recording UUID as `network_id`, not a Network model UUID); now also includes broadcasts where `metadata.user_id` matches the current user

---

### Bug fixes

| # | File | Issue | Fix |
|---|------|--------|-----|
| 1 | `DvrSchedulerService::matchManualRule()` | Dedup checked `scheduled_start = manual_start` but `scheduled_start` is stored as `manual_start − startEarlySeconds` — never matched, created a new recording every scheduler tick | Changed to `programme_start = manual_start` |
| 2 | `DvrSchedulerService::matchOnceRule()` | Once rules with `programme_id = null` silently did nothing — no way to schedule without real EPG data | When `programme_id` is null and the channel's playlist has dummy EPG enabled, falls back to the current dummy EPG slot (midnight-aligned, same algorithm as `EpgGenerateController`). Rule is disabled after scheduling to preserve "record once" semantics |
| 3 | `broadcast_manager.py::_monitor_process()` | When `StopDvrRecording` called `/broadcast/{uuid}/stop`, `_stopping = True` caused the monitor to skip the callback entirely — post-processing never triggered | DVR broadcasts now always send `programme_ended` even on intentional stop. Non-DVR network broadcasts still skip (unchanged behaviour) |
| 4 | `broadcast_manager.py::_monitor_process()` | An intentional stop kills FFmpeg via signal, producing a non-zero exit code — previously mapped to `broadcast_failed` | `clean_exit = exit_code == 0 or (self._stopping and dvr_mode)` |
| 5 | `DvrRecordingRuleResource` | `->visible()` closures compared `$get('type') === DvrRuleType::Manual->value` (string). When Filament loads an existing record, `$get('type')` returns the cast enum instance — comparison always `false`, manual date fields invisible on edit | Added `isRuleType()` helper that accepts both enum instance and backing string |

---

### Tests

- `DvrSchedulerServiceTest` — 3 new tests covering the dummy EPG Once-rule fallback (schedules correctly, skips when dummy EPG disabled, deduplicates on re-enable)
- `DvrRecorderServiceTest` — fixed `mockProxy()` return type (PHP intersection type `M3uProxyService&MockInterface` fails at runtime with Mockery)
- All 82 DVR-related tests pass

sparkison:
...gonna merge this thing!! It requires `v0.4.4` of the proxy to work correctly now

PR #1061, opened 2026-04-21
fix: single-item Fetch TMDB/TVDB IDs silently returns no result (#1054)

### LINKED ISSUE #1054: Bug: Fetch TMDB ID not same behavior between 3 dots vs Bulk action
Hello

I run with v0.11.28-dev

if i select a Serie/VOD which has no TMDB id, there 2 ways to fetch TMDB ID:
1. 3 vertical dots -> Fetch TMDB ID: **no result**!
2. Bulk action -> Fetch TMDB ID with overwrite metada: **Found**!

The first 1 does not work.

Image

```
{
"vodChannelIds": 
null,
"seriesIds": 
[
45460
],
"vodPlaylistId": 
null,
"seriesPlaylistId": 
null,
"allVodPlaylists": 
false,
"allSeriesPlaylists": 
false,
"overwriteExisting": 
false,
"user": 
null,
"isChunkJob": 
false,
"sendCompletionNotification": 
true
}
```

Image

Image

Image

```
{
"vodChannelIds": 
null,
"seriesIds": 
[
45460
],
"vodPlaylistId": 
null,
"seriesPlaylistId": 
null,
"allVodPlaylists": 
false,
"allSeriesPlaylists": 
false,
"overwriteExisting": 
true,
"user": 
{
"class": 
"App\Models\User",
"id": 
1,
"relations": 
[
],
"connection": 
"pgsql",
"collectionClass": 
null
},
"isChunkJob": 
false,
"sendCompletionNotification": 
true
}
```

Warbs816:
## Summary

Fixes #1054 — clicking the 3-dot "Fetch TMDB/TVDB IDs" action on a single Series or VOD row returns no result, while the bulk action with overwrite works for the same record.

Two things are compounding:

- `FetchTmdbIds::buildVodQuery` / `buildSeriesQuery` apply `->where('user_id', $this->user?->id)` on the legacy IDs branch. When the single-item action dispatches without a user, this becomes `WHERE user_id IS NULL` and filters every row out. The sibling playlist-id branches were already switched to `->when($this->user, …)` in 952a000c but the legacy IDs branch was missed.
- The single-item Filament actions also don't pass `user` or `overwriteExisting`. So even once the query is fixed, a record with `last_metadata_fetch` set but no ID is skipped by the per-item guards at `FetchTmdbIds.php:478` (VOD) / `:784` (series) — which is surprising when the user has explicitly clicked "fetch this one".

## Changes

- `app/Jobs/FetchTmdbIds.php` — use `->when($this->user, …)` on the legacy IDs branches, matching the sibling branches. This also benefits the import pipeline callers in `Models/Series.php` and `Models/Channel.php` which dispatch without a user.
- `app/Filament/Resources/Series/SeriesResource.php` and `app/Filament/Resources/Vods/VodResource.php` single-item actions — pass `user: auth()->user()` and `overwriteExisting: true`, and add the TMDB API key guard that the bulk actions already have.
- `tests/Feature/FetchTmdbIdsTest.php` — three regression tests covering the null-user dispatch path (VOD + series) and the retry-after-failed-attempt case.

## Test plan

- [x] 3-dot action on a series with `last_metadata_fetch` set returns "1 processed" instead of silently completing with 0
- [x] 3-dot action on a VOD with `last_metadata_fetch` set behaves the same
- [x] Bulk actions still honor the `overwrite_existing` toggle exactly as before
- [x] TMDB API key missing → 3-dot action shows the "TMDB API Key Required" notification (matching bulk)

### DIFF
--- a/app/Filament/Resources/Series/SeriesResource.php
+++ b/app/Filament/Resources/Series/SeriesResource.php
@@ -294,12 +294,25 @@ class SeriesResource extends Resource implements CopilotResource
                     ->modalDescription(__('Fetch TMDB, TVDB, and IMDB IDs for this series from The Movie Database.'))
                     ->modalSubmitActionLabel(__('Fetch IDs now'))
                     ->action(function ($record) {
+                        $settings = app(GeneralSettings::class);
+                        if (empty($settings->tmdb_api_key)) {
+                            Notification::make()
+                                ->danger()
+                                ->title(__('TMDB API Key Required'))
+                                ->body(__('Please configure your TMDB API key in Settings > TMDB before using this feature.'))
+                                ->duration(10000)
+                                ->send();
+
+                            return;
+                        }
+
                         app('Illuminate\Contracts\Bus\Dispatcher')
                             ->dispatch(new FetchTmdbIds(
-                                seriesIds: [$record->id]
+                                seriesIds: [$record->id],
+                                overwriteExisting: true,
+                                user: auth()->user(),
                             ));
-                    })
-                    ->after(function () {
+
                         Notification::make()
                             ->success()
                             ->title(__('TMDB Search Started'))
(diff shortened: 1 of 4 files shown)


sparkison (commented):
Looks good, thank you! 🙏

PR #1068, opened 2026-04-25
fix: prevent media server refresh notification spam on bulk STRM sync

Serph91P:
Previously, bulk-syncing N series/VOD channels via Filament actions dispatched N independent SyncSeriesStrmFiles / SyncVodStrmFiles jobs, each of which dispatched its own RefreshMediaServerLibraryJob, causing N HTTP refresh calls and N user notifications (e.g. 2178 notifications for a 2178-series sync).

Fix has two layers:

1. Bulk-aware sync jobs:
   - SyncSeriesStrmFiles: new optional series_ids parameter; when set, batching, cleanup and media server refresh are scoped to those IDs. Propagated through CheckSeriesStrmProgress.
   - SyncVodStrmFiles: same pattern with channel_ids, propagated through CheckVodStrmProgress.
   - All Filament bulk/group actions (Series, Categories, VodGroups) now dispatch a single job with the collected IDs instead of looping.

2. Belt-and-suspenders dedup at the refresh job:
   - RefreshMediaServerLibraryJob now implements ShouldBeUnique with uniqueId scoped per integration and uniqueFor=300s, so any burst of duplicate dispatches collapses to a single execution and a single notification per integration.

### DIFF
--- a/app/Filament/Resources/Categories/CategoryResource.php
+++ b/app/Filament/Resources/Categories/CategoryResource.php
@@ -233,12 +233,15 @@ class CategoryResource extends Resource implements CopilotResource
                     Action::make('sync')
                         ->label(__('Sync Series .strm files'))
                         ->action(function ($record) {
-                            foreach ($record->enabled_series as $series) {
-                                app('Illuminate\Contracts\Bus\Dispatcher')
-                                    ->dispatch(new SyncSeriesStrmFiles(
-                                        series: $series,
-                                    ));
+                            $seriesIds = $record->enabled_series->pluck('id')->all();
+                            if (empty($seriesIds)) {
+                                return;
                             }
+                            app('Illuminate\Contracts\Bus\Dispatcher')
+                                ->dispatch(new SyncSeriesStrmFiles(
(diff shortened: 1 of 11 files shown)


Serph91P:
@sparkison something else first, something I realized when testing the new profile for sports.

sparkison:
> @sparkison something else first, something I realized when testing the new profile for sports.

Good catch! I should have caught this earlier 🤦 reviewing now and I'll get it merged shortly...

Serph91P:
> > @sparkison something else first, something I realized when testing the new profile for sports.
> 
> Good catch! I should have caught this earlier 🤦 reviewing now and I'll get it merged shortly...

Don't be like that :wink:

sparkison:
added a couple extra scopes as I realized these jobs were running global cleanup against _all_ stream file sync settings in the system, not just the model(s) passed. So I added scoping for Playlist ID, Channel/Series ID, and the newly added Channel/Series Array added by this PR when they are set. This should prevent cross-contamination between users when multiple users present.

PR #1070, opened 2026-04-27
feat: Smart Channels (#330)

### LINKED ISSUE #330: "Smart Channel" to Auto-Select Best Quality Stream
Yes. For popular events like sports, there are often multiple streams available in my playlist. I always want to watch the best quality stream, but this requires me to manually check each one to find the highest bitrate, FPS, and resolution combination. This is a tedious and repetitive process.

I would love to have a feature to create a **"Smart Channel"** or **"Dynamic Channel"**. This would work as follows:
1.  **Create a Smart Channel:** The user creates a new virtual channel (e.g., "Best F1 Stream").
2.  **Set Filter Criteria:** The user defines rules for this channel, such as filtering all source channels by name (e.g., "Formula 1", "Sky F1 HD").
3.  **Define Quality Priority:** The user specifies the priority for selecting the best stream, for example:
    * Priority 1: **Resolution** (e.g., prefer 2160p > 1080p > 720p)
    * Priority 2: **Bitrate** (highest is best)
    * Priority 3: **FPS** (highest is best)
4.  **Automatic Selection:** The m3u-editor would then periodically scan the filtered source streams and automatically link the "Smart Channel" to the URL of the stream that currently best matches the defined quality criteria.
5.  **(Optional) Automatic Failover:** If the best stream goes offline, the Smart Channel should automatically switch to the second-best available stream.

This would ensure that when I tune into my "Best F1 Stream", I'm always watching the highest quality version available without any manual intervention. A possible technical approach for analyzing stream quality could involve using `ffmpeg`, as demonstrated in projects like [IPTVChecker-BitRate](https://github.com/sudo-ronmexico/IPTVChecker-BitRate).

Warbs816:
Closes #330.

## Summary

Adds Smart Channels — custom channels that automatically route to the highest-quality source from a configured pool of failovers, with the ranking kept current via periodic and manual rescoring against actual stream stats.

The implementation reuses the existing custom-channel-with-empty-URL fallback in `PlaylistUrlService::getChannelUrl()`: a smart channel has no URL of its own, and the playlist streams whatever sits at `channel_failovers.sort = 0`. Re-scoring just rewrites the sort column — the master is never promoted, swapped, or modified, so the user-facing identity (title, logo, EPG mapping) is stable forever.

## What's new

**Scoring extensions**
- Frame Rate and Bitrate added as priority attributes alongside the existing Resolution/Codec/etc. Scoring rules are normalized to 0–100.
- Probe enrichment: ffprobe now also requests `-show_format`. When ffprobe leaves both per-stream and format-level video bitrate null (typical for live MPEG-TS / HLS), a short follow-up packet-sampling probe (`-read_intervals \"%+5\"`) measures actual throughput and back-fills it. Tested live against MPEG-TS and HLS sources.

**Smart channel creation**
- "Make smart channel" bulk action on the channels resource. Pre-creation modal shows a ranked preview with per-attribute breakdowns so users can see why the order is what it is before committing.
- New `is_smart_channel` flag on channels (boolean column with index, backfill migration). Used for query scoping, UI gating, list-page badges, edit-form lockdown of the URL field, and the Stream Monitor badge when an active stream is backed by a smart channel.

**Failover rescoring**
- New `RescoreChannelFailovers` job. Iterates failover groups owned by a playlist, re-probes stale members (within the configured staleness window, throttled via `ProviderRequestDelay`), and updates `channel_failovers.sort`. Master is never altered.
- Console command `app:rescore-channel-failovers` runs hourly and dispatches per-playlist when the configured interval has elapsed. Single-playlist `--playlist <id>` invocation also works for ad-hoc runs.
- `WithoutOverlapping` queue middleware keyed on `playlistId` so manual + scheduled fires can't double up against the same playlist.
- Per-channel "Rescore now" button on the channel info pane.

**Visibility**
- Failover Ranking section in the channel info pane: expandable cards per failover (rank + score + breakdown badges), with native `<details>` disclosures revealing each failover's technical details (resolution, fps, bitrate, codec, audio info, last probed). Score breakdowns persist to `channel_failovers.metadata` so they're durable, not just a one-shot modal preview.
- Smart Channel column in the channels list (sparkles icon, info color, only renders for flagged rows).
- Smart Channel badge on the Stream Monitor when a smart channel's stream is in flight.

**Per-playlist settings**
- New \"Failover Rescoring\" section on the playlist edit page (always visible). Houses the schedule (Off / Daily / Weekly) and the staleness window in days (default 7). The scoring rules themselves come from the existing Auto-Merge \"Priority Order\" — the new section just controls when and how often things rescore.

## How it works under the hood

A smart channel is a custom channel where:

- `is_custom = true`
- `is_smart_channel = true` (new column)
- `url` is null
- One or more `channel_failovers` rows attach source channels, sorted by score.

`PlaylistUrlService::getChannelUrl()` already had the fallback logic for the first two conditions, so the streaming path needed zero changes. Everything new is around scoring, ranking, and surfacing the rationale.

## Known limitations / deferred

- **Channel scrubber doesn't know about failover relationships** — a scrubber rule can disable a smart channel's source channels and silently degrade the smart channel's quality. Worth a follow-up issue with a discussion about how scrubbers should treat dependents.
- **HLS variant selection is whatever ffprobe picks** — for multi-variant HLS sources, the score reflects the variant ffprobe selected (usually the highest bandwidth listed), not necessarily the variant the user's player would actually adapt to. Could be addressed later with explicit variant-selection flags or per-variant probing.
- **\"Make smart channel\" bulk action uses existing stats only** — it doesn't re-probe stale sources before scoring. Users wanting fresh stats can probe the sources first, or rescore the smart channel right after creation. Documented in the playlist settings helper text.

## Test plan

- [x] Probe a non-smart MPEG-TS live channel and confirm Technical Details now shows a video bitrate (was previously null for many streams)
- [x] Probe an HLS channel and confirm the same
- [x] In a playlist's auto-merge config, add Frame Rate or Bitrate to the Priority Order, run a sync, and confirm the master selection respects them
- [x] On the channels list, multi-select 2+ source channels in the same playlist → \"Make smart channel\" → check the modal preview shows the ranked breakdown → confirm → verify the new smart channel appears with the sparkles badge
- [x] Verify the smart channel streams correctly via the M3U / HDHR / Xtream URL — should serve from the top failover
- [x] Open the smart channel's info pane: Technical Details should be hidden; Failover Ranking section visible with expandable cards showing per-failover stats
- [x] Click \"Rescore now\" on the info pane → verify a fresh ranking after the queue worker picks up the job
- [x] On the playlist edit page, set the Failover Rescoring interval to Daily → verify `app:rescore-channel-failovers` dispatches a job when run with no args
- [x] Open the Stream Monitor while the smart channel is being streamed → confirm the Smart Channel badge appears
- [ ] Try \"Make smart channel\" with channels from two different playlists → expect a warning notification, no creation
- [x] Try attaching an existing smart channel as a failover via the per-channel Repeater search → expect it to not appear in results


sparkison:
thanks for the effort on this, and keeping it going! Unfortunately I've still not had to test or review so I'm going to close this down for now rather than just keeping it open hoping I'll have the time to get to it to review and test. Thanks again, and sorry for leaving it hang so long ☹️

PR #1080, opened 2026-04-30
feat: add TRaSH Guides naming support for VOD and series

Serph91P:
- Add VodFileNameService for Trash-Guide-conformant movie filenames
- Add SerieFileNameService for SxxExx episode filenames
- Add VersionDetectionService for edition tag detection
- Add StrmPathBuilder for centralized path building
- Extend StreamFileSettingResource with Trash Guide Naming fieldset
- Update SyncVodStrmFiles and SyncSeriesStrmFiles jobs
- Fix Episode model stream_stats and release_group accessors
- Add migrations for trash guide fields


Serph91P:
@sparkison hope that this is ok?

sparkison:
> @sparkison hope that this is ok?

I'm not familiar with TRaSH-Guides, other than a quick search just now. What does this add specifically? Just trying to understand a bit more what we're trying to implement here 😉

Serph91P:
> > @sparkison hope that this is ok?
> 
> I'm not familiar with TRaSH-Guides, other than a quick search just now. What does this add specifically? Just trying to understand a bit more what we're trying to implement here 😉

They basically suggested a folder and file naming scheme for all media servers to pick up the information from the path and file names. And I tried to mimic this for the best compatibility with media servers so they can recognize multiple versions of the same movie and so on.

See here for example:
https://trash-guides.info/Sonarr/Sonarr-recommended-naming-scheme/

Or here:
https://trash-guides.info/Radarr/Radarr-recommended-naming-scheme/

sparkison:
got it! Thanks for that, definitely seeing some issues though. For one, I don't see where the `stream_stats` is actually populated for Episodes? Same for the `release_group` for channel and episodes? These feel like empty columns that are never populated, so not sure why they're added?

Other items to address:

### Critical / Blocking Issues

**1. Opt-out is impossible — existing users will get trash guide naming forced on them**

The `movie_format` column is added with a non-null default value:

```sql
DEFAULT '{title} ({year}){edition}{quality}{audio}{video}{-group}'
```

This means every existing `StreamFileSetting` record will have `movie_format` populated after the migration. In `SyncVodStrmFiles`:

```php
$trashGuideEnabled = filled($streamFileSetting?->movie_format ?? null);
```

This will be `true` for **all** existing settings — even ones created before this PR. Every VOD sync will switch to trash guide naming without any user action. This is a silent breaking change. There should be either a `trash_guide_naming` boolean toggle (defaulting to `false`), or `movie_format`/`episode_format` should default to `null`.

**2. `{-group}` placeholder is missing from `VodFileNameService`**

The default `movie_format` (in both the migration and the Filament form) is:

```
{title} ({year}){edition}{quality}{audio}{video}{-group}
```

But `VodFileNameService::generateMovieFileName()` only maps `{group}`, not `{-group}`. The `cleanUnfilledPlaceholders()` method strips unrecognized placeholders, so `{-group}` will always be silently removed — even when the channel has a release group set. `SerieFileNameService` correctly implements `{-group}`. This needs to be added to `VodFileNameService`.

**3. Legacy VOD title-folder creation removed (regression)**

In the old `SyncVodStrmFiles`, when `title` was in `pathStructure`, the code created a title sub-folder with `mkdir` and updated `$path`. In the new code's `else` (legacy) branch, this folder creation is gone entirely. Existing users with `path_structure: ['group', 'title']` who aren't using trash guide naming will have their files written flat into the group folder instead of into per-title sub-folders.

---

### High Priority Issues

**4. `{-ep_title}` helper text doesn't match the actual placeholder**

The Filament form advertises `{-ep_title}` in its helper text, but `SerieFileNameService` implements `{-title}` (a prefix-dash variant of `{ep_title}`). The placeholder `{-ep_title}` doesn't exist. This will confuse users and silently produce wrong filenames.

**5. Unsafe user-supplied regex execution**

In `VersionDetectionService::detectEditionWithPattern()`:

```php
if (preg_match($pattern, $title, $matches) === 1) {
```

`$pattern` comes directly from the user-editable `version_detection_pattern` field. An invalid regex will produce a PHP warning/error. Wrap this in `try/catch` or validate the pattern before saving and before calling `preg_match`.

**6. `Series::episodes()` changed to `hasManyThrough` — potentially breaking**

The change from `hasMany(Episode::class)` (via direct `series_id`) to `hasManyThrough(Episode::class, Season::class)` means episodes without a season won't appear via this relationship. The existing `enabled_episodes()` method still uses `hasMany`, creating an inconsistency. Any code or query that loads `series.episodes` may behave differently. Needs careful review of all call sites.

**7. All three migrations share the same timestamp**

```
2026_04_30_092713_add_trash_guide_fields_to_channels_table.php
2026_04_30_092713_add_trash_guide_fields_to_stream_file_settings_table.php
2026_04_30_092713_update_episodes_for_trash_guide_naming.php
```

Laravel resolves tie-breaking alphabetically, but this is fragile and can trip up tooling. The timestamps should be sequential (e.g., `_092713`, `_092714`, `_092715`).

---

### Medium Priority Issues

**8. `StrmPathBuilder::buildVodPath` calls `generateMovieFileName` twice**

```php
// Called once to build the title folder name:
$fileName = $this->vodFileNameService->generateMovieFileName($channel, $setting);
$folderName = $this->buildVodTitleFolder($channel, $fileName, $syncSettings);

// Then called again for the actual file:
$fileName = $this->vodFileNameService->generateMovieFileName($channel, $setting);
```

The service isn't cached; it does regex, string normalization, etc. Store the result.

**9. Inconsistent quality thresholds between the two services**

`VodFileNameService` uses `height >= 1000` for `1080p` and `height >= 700` for `720p`, while `SerieFileNameService` uses the more standard `height >= 1080` and `height >= 720`. A 1000px-tall source would be labeled `1080p` by the VOD service. These should share the same logic, ideally in one place.

**10. Duplicate stream stats normalization logic**

Both `VodFileNameService::normalizeStreamStats()` and `SerieFileNameService::normaliseStreamStats()` do the same job — flattening the nested ffprobe structure — with subtly different implementations (one returns a flat map, one returns `{video, audio, flat}`). This should be one shared method/trait/class.

**11. Stray `use PDO;` import in `AppServiceProvider`**

`PDO` appears to be imported but not used in the visible diff. Should be removed.

---

### Summary

The feature itself is well-conceived and mostly well-implemented, but **issues 1 and 3 are regressions that would affect all existing users on first sync after upgrading**. Issue 2 means the default format the PR ships with doesn't work as documented. I'd recommend addressing those three before merging.

Serph91P:
@sparkison stap I am already doing stuff :wink: and yes I just realized that I pushed my sport test stuff sorry about that :smile:

sparkison:
> @sparkison stap I am already doing stuff 😉 and yes I just realized that I pushed my sport test stuff sorry about that 😄

lol! 😂 sorry about that, thought it was ready for review.

Serph91P:
@sparkison I think I got it now, can you please check again. I will implement one last thing, that there is a third state for the probed channels, so it says "not probed", "probed" and "failed" or something when the probe did not return any result or failed or whatever.

PR #1081, opened 2026-04-30
feat: Bulk-assign stream profile to channels and groups

Warbs816:
## Summary

Adds a **Set Stream Profile** action so a profile can be applied to many records in one step rather than opening each channel's edit pane individually.

**Where it lives:**
- Channel + VOD bulk modals — under the existing **Streaming** section
- Live + VOD group bulk toolbar
- Live + VOD group per-row (three-dots) menu

**Form:**
- `Stream Profile` — searchable select; pick `None` to clear
- `Overwrite existing assignments` — default **off**. When off, only channels with a null profile are touched; already-set channels (e.g. a hand-tuned BBC One) are preserved. When on, all selected channels are reassigned (or cleared if `None` is picked).
- Group actions add `Apply to channels added later` — saves the profile on the group itself so new channels imported into the group inherit it automatically.

**Auto-inheritance for new channels:**
- `ChannelObserver::creating()` covers `Channel::create()` / `updateOrCreate()` paths.
- `ProcessM3uImportChunk` and `ProcessM3uVodImportChunk` inject the group's `stream_profile_id` into their bulk payload, since `Channel::upsert()` bypasses model events.
- The upsert's `update:` list still excludes `stream_profile_id`, so re-imports never overwrite existing per-channel assignments.

**Schema change:** `groups.stream_profile_id` (nullable FK, `nullOnDelete`).

## Test plan

- [x] On the **Live Channels** list, multi-select a few channels with mixed profile state; run the bulk **Set Stream Profile** action with overwrite off — only previously-null channels should change.
- [x] Repeat with overwrite on — every selected channel switches to the chosen profile (including being cleared when `None` is picked).
- [x] Same on the **VOD** list.
- [x] On the **Live Groups** page, three-dots a single group → **Set Stream Profile** with `Apply to channels added later` on; confirm the group default sticks and existing live channels update.
- [x] Repeat from the bulk toolbar across multiple selected live groups; confirm only `live_channels` are touched (a VOD channel that happens to share the group is left alone).
- [x] Same flow on **VOD Groups**, mirrored.
- [ ] After setting a group default, run a playlist re-sync; newly imported channels in that group inherit the default while existing channels keep their previously-set profile.

### DIFF
--- a/app/Filament/Resources/Channels/ChannelResource.php
+++ b/app/Filament/Resources/Channels/ChannelResource.php
@@ -978,6 +978,47 @@ class ChannelResource extends Resource implements CopilotResource
 
             // -- Streaming --
             BulkModalActionGroup::section('Streaming', [
+                BulkAction::make('set-stream-profile')
+                    ->label(__('Set Stream Profile'))
+                    ->schema([
+                        Select::make('stream_profile_id')
+                            ->label(__('Stream Profile'))
+                            ->options(fn () => StreamProfile::where('user_id', auth()->id())->pluck('name', 'id'))
+                            ->searchable()
+                            ->preload()
+                            ->nullable()
+                            ->placeholder(__('None (clear profile)'))
+                            ->helperText(__('The stored profile only takes effect when the channel (or its playlist) has proxy enabled.')),
+                        Toggle::make('overwrite_existing')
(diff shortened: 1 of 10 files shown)


sparkison:
Looks good! Gonna get it merged for the next release 😉

PR #1088, opened 2026-05-02
feat: Show live media info and failover channel name on Stream Monitor

### LINKED ISSUE #1085: Stream Monitor to display detailed properties of video being streamed

Feature request is not related to a specific problem other than a desire for more visibility into what's being streamed to my clients.

The Stream Monitor currently lacks any information about the video being streamed to clients. Information I'd like to see would include video resolution, codec, audio properties and anything else (transcoder in use? remote stream connection being shared between multiple clients?) we can pick out from the video stream. The information should be displayed in a relatively compact fashion since it would have to be displayed per client connection.

Dispatcharr provides something like this:

Warbs816:
## Summary

Closes #1085. Stream Monitor now shows live media badges and the active failover channel name, matching the Dispatcharr-style readout requested in the issue.

Pairs with the proxy-side PR that publishes the data: m3ue/m3u-proxy#54.

## What changed

**Failover channel name** — when a stream is on a failover URL the title renders as `Primary Name → Failover Name`, with the failover name in orange to match the existing *Failover Active* badge styling. The active failover is identified by URL-matching `current_url` against `PlaylistUrlService::getChannelUrl()` for each candidate, which means it works correctly under the dynamic resolver (where the proxy's `current_failover_index` doesn't necessarily line up with the candidate's slot in `failoverChannels`). Falls back to index lookup for the static-list mode.

**Live media badges** — the page now shows resolution / fps / codec / audio / container / bitrate / speed when the proxy reports them. Badges are conditional on the proxy returning a non-empty `media_info`, so plain HTTP-proxy streams just don't show them rather than rendering empty placeholders.

The page only consumes data the proxy already publishes — there's no extra polling or work in the editor.

## Test plan

- [x] Open Stream Monitor while a transcoded stream is active — resolution, codec, fps and bitrate badges populate within a few seconds
- [x] Plain HTTP proxy stream — no badges (clean, not empty placeholders)
- [x] Trigger failover — title shows `Primary → Failover` with the failover name in orange; badges repopulate for the failover stream
- [x] With dynamic failover resolver enabled and the first candidate skipped, the title resolves to the *actually used* candidate (not the first one in the list)

### DIFF
--- a/app/Filament/Pages/M3uProxyStreamMonitor.php
+++ b/app/Filament/Pages/M3uProxyStreamMonitor.php
@@ -3,6 +3,7 @@
 namespace App\Filament\Pages;
 
 use App\Facades\LogoFacade;
+use App\Facades\PlaylistFacade;
 use App\Models\Channel;
 use App\Models\Episode;
 use App\Models\Network;
@@ -11,6 +12,7 @@ use App\Models\PlaylistAlias;
 use App\Models\PlaylistProfile;
 use App\Models\StreamProfile;
 use App\Services\M3uProxyService;
+use App\Services\PlaylistUrlService;
 use Carbon\Carbon;
 use Exception;
 use Filament\Actions\Action;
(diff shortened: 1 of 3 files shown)


Warbs816:
@sparkison let me know how to proceed as per comment on https://github.com/m3ue/m3u-editor/issues/1085!

Warbs816:
Superseded by #1089, which rebases this work onto dev's new badge UI from 0146613b and layers the live proxy media_info on top of the stored probe data instead of replacing it.

PR #1089, opened 2026-05-02
feat: Layer live proxy media info on Stream Monitor + failover channel name

### LINKED ISSUE #1085: Stream Monitor to display detailed properties of video being streamed

Warbs816:
## Summary

Builds on the Stream Monitor media badges added in 0146613b. Layers two things on top:

1. **Live ffmpeg media info** from the proxy overrides the stored ffprobe data per-field, so badges reflect what's actually being streamed *right now* rather than what was probed at sync time. Probe data still fills in fields the proxy doesn't expose (video profile, audio language, audio bitrate) and remains the sole source for plain HTTP-proxy streams that don't transcode.
2. **Failover channel name** in the title — when on a failover URL the title renders as `Primary → Failover` with the failover name in orange, matching the existing Failover Active badge.

Closes #1085. Pairs with the proxy-side PR that publishes the live data: m3ue/m3u-proxy#54.

Supersedes #1088 (which was based on the previous dev HEAD before badges had landed there).

## What changed

**Live data layering** (`app/Filament/Pages/M3uProxyStreamMonitor.php`)

The existing block that pulls `getEmbyStreamStats()` for `media_info` now also reads `media_info` from the proxy stream payload. Proxy keys are mapped to the editor's existing field shape:

| Proxy field | Editor field |
| --- | --- |
| `resolution` | `resolution` |
| `video_codec` | `video_codec` |
| `fps` | `source_fps` |
| `bitrate_kbps` | `video_bitrate_kbps` |
| `audio_codec` | `audio_codec` |
| `audio_channels` | `audio_channels` |

`array_filter` strips empty live values so the merge only overrides where the proxy actually has data, keeping `video_profile`, `audio_language`, and `audio_bitrate_kbps` from the probe path. The blade renders unchanged keys, so no UI rework was needed.

**Failover channel resolution**

The active failover is identified by URL-matching `current_url` against `PlaylistUrlService::getChannelUrl()` for each candidate in `failoverChannels`. Static-list mode falls back to index lookup if the URL match fails. The dynamic resolver case (where the resolver skipped earlier candidates and `current_failover_index` doesn't line up with the candidate's slot) is the reason for the URL-match path — covered by a dedicated test.

`failoverChannels` is eager-loaded on the channel pre-fetch to keep this from regressing the page's N+1 budget.

**Title display** (blade)

Drops a `Primary → Failover` line into the existing title block when `failover_channel.title` is populated. Reuses the existing orange used by the Failover Active badge.

## Test plan

- [x] Open Stream Monitor with a transcoded stream that has stored probe data — badges populate immediately from probe, then live values overlay on resolution/fps/bitrate within ~1-3s
- [x] Plain HTTP proxy stream — badges still show from probe data only (no live data path)
- [x] Stream with no probe data and no transcoding — no badges (as before)
- [x] Trigger failover — title shows `Primary → Failover` in orange; badges repopulate from the failover stream's live data once the new ffmpeg starts emitting Stream lines
- [x] Dynamic failover resolver with the first candidate skipped — title resolves to the *actually used* candidate, not the first one in the list

### DIFF
--- a/app/Filament/Pages/M3uProxyStreamMonitor.php
+++ b/app/Filament/Pages/M3uProxyStreamMonitor.php
@@ -11,6 +11,7 @@ use App\Models\PlaylistAlias;
 use App\Models\PlaylistProfile;
 use App\Models\StreamProfile;
 use App\Services\M3uProxyService;
+use App\Services\PlaylistUrlService;
 use Carbon\Carbon;
 use Exception;
 use Filament\Actions\Action;
@@ -297,7 +298,10 @@ class M3uProxyStreamMonitor extends Page
                 ->values();
 
             $channelsById = $channelIds->isNotEmpty()
-                ? Channel::whereIn('id', $channelIds)->get()->keyBy('id')
+                ? Channel::whereIn('id', $channelIds)
+                    ->with('failoverChannels')
+                    ->get()
+                    ->keyBy('id')
                 : collect();
             $episodesById = $episodeIds->isNotEmpty()
                 ? Episode::whereIn('id', $episodeIds)->get()->keyBy('id')
@@ -330,6 +334,7 @@ class M3uProxyStreamMonitor extends Page
                 $model = [];
                 $title = null;
                 $logo = null;
+                $failoverChannel = null;
                 if (isset($stream['metadata']['type']) && isset($stream['metadata']['id'])) {
                     $modelType = $stream['metadata']['type'];
                     $modelId = $stream['metadata']['id'];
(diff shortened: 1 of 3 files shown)


sparkison:
Looks great!

PR #1094, opened 2026-05-04
feat(channel): derive video bitrate from container for live MPEG-TS streams

Serph91P:
ffprobe does not expose a per-stream `bit_rate` on the video elementary stream of live MPEG-TS feeds (no CBR container, unknown duration), so `getEmbyStreamStats()` returned `ffmpeg_output_bitrate = null`. Downstream consumers (e.g. emby-xtream) then fell back to the audio bitrate as the MediaSource bitrate, which is wildly wrong (e.g. 384 kbps for a 1080p mux).

This is the cheapest fix that gets a useful value:

- Add `-show_format` to the existing ffprobe call (one extra section, no extra round-trip, no extra timeout).
- Persist the container `format.bit_rate` alongside the streams in `stream_stats` as a `['format' => ['bit_rate' => …]]` entry (kept separate from stream entries so existing consumers still iterate streams unchanged).
- In `getEmbyStreamStats()`, prefer the per-stream video `bit_rate` when present; otherwise fall back to `format.bit_rate - audio.bit_rate`, which is a tight upper bound for the video bitrate on a typical 1 video + 1 audio TS mux.

Tests cover all three paths: live TS fallback, per-stream value wins when both are available, and null when neither is available.

### DIFF
--- a/app/Models/Channel.php
+++ b/app/Models/Channel.php
@@ -367,7 +367,7 @@ class Channel extends Model
                 return [];
             }
 
-            $process = new SymfonyProcess(['ffprobe', '-v', 'quiet', '-print_format', 'json', '-show_streams', $url]);
+            $process = new SymfonyProcess(['ffprobe', '-v', 'quiet', '-print_format', 'json', '-show_streams', '-show_format', $url]);
             $process->setTimeout($timeout);
             $process->run();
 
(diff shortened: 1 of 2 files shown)


Serph91P:
@sparkison small fix to actually fill the bitrate for ts streams :wink:

PR #1095, opened 2026-05-04
feat: Render live encoder output stats on Stream Monitor

Warbs816:
## Summary

Pairs with the proxy-side change that exposes `output_media_info` — m3ue/m3u-proxy#56. The existing Stream Info row reflects the source ffmpeg is reading; this PR adds an Output row beneath it showing what ffmpeg is *producing* (encoder codec, target resolution, muxer container, fps, bitrate, audio).

Closes the question raised on m3ue/m3u-editor#1089 about layering output-side stats too.

## What changed

**Model mapping** (`app/Filament/Pages/M3uProxyStreamMonitor.php`)

After the existing live `media_info` merge, we read `$stream['output_media_info']` from the proxy payload and attach a filtered copy as `$model['output_media_info']` when non-empty. Mirrors the existing `media_info` mapping shape — no probe-data fallback because output stats only exist when ffmpeg is actively transcoding (no encoder = no row).

**Blade row** (`resources/views/filament/pages/m3u-proxy-stream-monitor.blade.php`)

Adds a new Output Info row directly below the Stream Info row, gated on `$stream['model']['output_media_info']`. Container palette is violet (`bg-violet-50` / `dark:bg-violet-950/30`) to make it visually distinct from the failover orange used elsewhere on the card. Per-field badge colours match the Stream Info row (`info` for resolution, `primary` for codec, `success` for audio, `gray` for fps/bitrate/container) so input ↔ output is comparable field-by-field. Includes the same green pulse dot used on the input row to indicate live data.

## Behaviour

- Transcoded stream: Stream Info shows source description; Output shows what ffmpeg is encoding to right now (e.g. source 1920×1080 hevc → output 1280×720 h264)
- Plain HTTP-proxy stream: no Output row (proxy reports no `output_media_info`)
- Resolver stream (yt-dlp / streamlink): no Output row

## Test plan

- [x] Transcoded stream — Stream Info row still populated from probe + live data; Output row appears with encoder fields
- [x] Plain HTTP-proxy stream — Stream Info still works (probe data only); Output row absent
- [x] Resolver stream — Both rows absent
- [x] Choose a better colour?


Warbs816:
Can't decide on the colour for the output data bg... I'll let you decide 😊

sparkison:
Reviewing and merging shortly!

Issue #1105, opened 2026-05-08
Bug: Movie extra components -> Selection not working


fabien4444:
Hello

i run with v0.11.35-dev

I want to append edition + some infos among quality/codec/audio/HDR 

For that, i set in Stream File Settings -> Enable Trash Guide naming (extra components)

But it takes all infos in the name of file, not my selection!

Here example: i want only Edition + HDR but in the name of file i have quality/codec/audio/HDR 

<img width="477" height="226" alt="Image" src="https://github.com/user-attachments/assets/a7348859-402b-4867-9d09-72714eb3e091" />

<img width="390" height="28" alt="Image" src="https://github.com/user-attachments/assets/4a7c2f87-a7e8-4d6b-adbe-d4bf30c94d6e" />

Serph91P:
@fabien4444 have you changed the selection after creation?

fabien4444:
Yes

Serph91P:
> Yes

Have you rerun the processing or recreated the STRM files?

fabien4444:
Bulk action on a VOD -> Sync STRM files

Serph91P:
> Bulk action on a VOD -> Sync STRM files

Sorry, but is this an answer or a question?

fabien4444:
Hello

here the steps i did:
1. Create a Stream File Settings for VOD with full extra infos selected
2. Select a VOD and 'Sync VOD Strm files'

<img width="321" height="263" alt="Image" src="https://github.com/user-attachments/assets/fce12af8-ac1e-4be0-849c-b1115625e4a7" />

3. The result:

<img width="483" height="21" alt="Image" src="https://github.com/user-attachments/assets/9fd7a128-89dd-458a-a7de-889cbae5901c" />

4. Edit the Stream File Settings to select only 'Edition' & 'HDR'
5. Did step 2 again and same result step 2

Serph91P:
@fabien4444 looks like this now and is working: 
```
'Marco 2026 [H.265 E-AC-3 5.1].nfo'  'Marco 2026 [H.265 E-AC-3 5.1].strm'
```

fabien4444:
Hello

it's better but HDR seems not working!
I select only Quality and HDR but only quality is in the filename

Serph91P:
> Hello
> 
> it's better but HDR seems not working! I select only Quality and HDR but only quality is in the filename

Did you probe the channels? Else, it cannot know the other indicators. Like I have shown you it works for me 😄

[...]

PR #1106, opened 2026-05-08
fix(strm): apply trash-guide naming additively to legacy filenames

Serph91P:
Previously enabling trash_guide_naming_enabled bypassed the legacy filename builder via StrmPathBuilder, which meant year, tmdb_id, group, and other filename_metadata options were silently dropped from .strm filenames.

Both SyncVodStrmFiles and SyncSeriesStrmFiles now run a single code path: build the legacy filename, then append the trash-guide extras bracket additively (after year, before tmdb id and group). Removes the StrmPathBuilder import from both jobs.

Provider-driven only:
- VodFileNameService::resolveEdition no longer guesses the edition from title/name (a token like "4K UHD" is a quality marker, not an edition); only explicit channel.edition or info/movie_data edition fields are honoured.
- Quality / audio / hdr no longer fall back to title parsing; only probed stream_stats or an explicit manual value are used. If stream_stats are missing the bracket is omitted entirely.

Tested: 34 trash-guide / strm-naming tests pass.


Serph91P:
@Contributors @sparkison if someone could test this, please, to confirm that this is working.

sparkison:
> @contributors @sparkison if someone could test this, please, to confirm that this is working.

Looks good to me! Removed the unused class since this is additive now and it wasn't used (the older trash guide path builder). Otherwise, testing a bunch of different options and seemed to be working as expected!

Serph91P:
> > @contributors @sparkison if someone could test this, please, to confirm that this is working.
> 
> Looks good to me! Removed the unused class since this is additive now and it wasn't used (the older trash guide path builder). Otherwise, testing a bunch of different options and seemed to be working as expected!

Thanks a lot. Then we finally can close this topic https://github.com/m3ue/m3u-editor/issues/1105

PR #1108, opened 2026-05-09
fix(probe): aggregate manual bulk-probe completion via Bus::batch

Serph91P:
The bulk "Probe Streams" actions in VodResource, SeriesResource and the EpisodesRelationManager dispatched ProbeVodStreamsChunk jobs in a loop and asked only the last-dispatched chunk to send the completion notification with the total count.

This was wrong on two counts:
1. Loose chunks have no ordering guarantee. The last-dispatched chunk often finished first, firing a notification like "3 of 156 probed successfully" while 100+ records were still being processed.
2. Even when chunks finished in order, the per-chunk counter only reflected that final chunk's work, not the whole batch.

Fix: introduce ProbeManualComplete, a dedicated completion job dispatched through Bus::batch()->finally(). It records the start timestamp and, when the entire batch finishes, counts how many of the originally selected channels/episodes have stream_stats_probed_at >= start. That is the real "probed successfully" number and the user only sees one accurate notification at the end.

Refactor target call sites:
- app/Filament/Resources/Vods/VodResource.php (bulk action)
- app/Filament/Resources/Series/SeriesResource.php (single-record series-probe action and bulk action)
- app/Filament/Resources/Series/RelationManagers/EpisodesRelationManager.php (bulk action)

Single-record probe actions on individual VOD channels (VodResource line 670) keep dispatching ProbeVodStreamsChunk directly, because a single chunk has no ordering bug.

The job_batches table is already provisioned (see the existing remove_batch_job_tables migration which only truncates, never drops it), and Bus::batch is already in use elsewhere in the probe subsystem (ProbeChannelStreams, ProbeVodStreams), so no infra changes are needed.

### DIFF
--- a/app/Filament/Resources/Series/RelationManagers/EpisodesRelationManager.php
+++ b/app/Filament/Resources/Series/RelationManagers/EpisodesRelationManager.php
@@ -3,7 +3,7 @@
 namespace App\Filament\Resources\Series\RelationManagers;
 
 use App\Filament\Tables\ProbeStatusColumn;
-use App\Jobs\ProbeVodStreamsChunk;
+use App\Jobs\ProbeManualComplete;
 use Filament\Actions;
 use Filament\Actions\Action;
 use Filament\Actions\ViewAction;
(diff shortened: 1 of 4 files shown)


sparkison:
Good catch! Going to run into issues batching with the provider request delay though. Went with something a little simpler using chain instead for enforced order 😉

PR #1111, opened 2026-05-09
DVR: From Original Source channel, seriesNewOnly fix, guest series widget, playlist_auth_id propagation

Grimothy:
## Summary

- **From Original Source** channel option for series rules — when browsing a show, the channel you're browsing is stored as `source_channel_id` on the rule; the scheduler uses it to scope EPG lookup and stream URL, giving priority to the exact channel without locking the rule to that channel
- **`seriesNewOnly` bool→int fix** — Livewire 4 serialises PHP `bool` as a JS boolean, which never matches HTML `<option value="0">` string values, causing the select to snap back on every interaction; changed to `public int $seriesNewOnly = 0`
- **`GuestScheduledSeriesWidget`** — header widget on the guest Recordings list that shows all active series rules, giving guests immediate feedback after scheduling (actual `DvrRecording` rows only appear within 30 min of episode start)
- **`playlist_auth_id` propagation** — copied from the recording rule onto the created `DvrRecording` in both `createScheduledRecordingFromProgramme` and `matchManualRule`; fixes the guest Cancel button which was permanently hidden because `record->playlist_auth_id` was `null`vv
- **fix: don't append APP_PORT to HTTPS URLs in console base URL**
When APP_URL uses https:// (i.e. the app is behind a reverse proxy), configureConsoleBaseUrl() was unconditionally appending APP_PORT to the URL, producing incorrect external URLs like https://host:36400 in all queue worker contexts (Horizon jobs, Artisan commands).

This caused media server integration channels (Emby/Jellyfin) to have the internal container port baked into their stored stream/logo URLs, surfacing in the Channel URL Settings panel as port-appended provider URLs.

The fix mirrors the existing guard in PlaylistService::getBaseUrl(): only append the port for http:// URLs. https:// implies a reverse proxy terminating TLS at port 443 so APP_PORT is not the external port.

## Changes

- Migration: `source_channel_id` nullable FK on `dvr_recording_rules` (nullOnDelete)
- `DvrRecordingRule`: `source_channel_id` fillable + `sourceChannel(): BelongsTo`
- `DvrSchedulerService`: priority 2.5 branch in `resolveSeriesEpgScope` + `resolveStreamUrl` for `source_channel_id`; eager-loads `sourceChannel.epgChannel`
- `BrowseShows` + `GuestBrowseShows`: `$sourceChannelId ?int`; sentinel `seriesChannelId = 0` pre-populated from browsed channel; resolved to `null + source_channel_id` before persisting
- `browse-show-detail.blade.php`: "From Original Source — {channel}" option above "Any channel"
- `DvrRecordingRuleResource` + `GuestDvrRuleResource`: channel select prepends `0 → 'From Original Source'`; EditAction sentinel round-trip via `mutateRecordDataUsing`/`mutateDataUsing`
- All 100 tests pass (49 DvrScheduler + 51 BrowseShows)

### DIFF
--- a/app/Filament/GuestPanel/Pages/GuestBrowseShows.php
+++ b/app/Filament/GuestPanel/Pages/GuestBrowseShows.php
@@ -88,9 +88,13 @@ class GuestBrowseShows extends Page
 
     // --- Series options form state ---
 
-    public bool $seriesNewOnly = false;
+    public int $seriesNewOnly = 0;
 
-    public ?int $seriesChannelId = null;
+    /** 0 = "From Original Source" (sentinel), null/negative = "Any channel", int = specific channel */
+    public int $seriesChannelId = 0;
+
+    /** The actual channel ID the show was browsed from (null if not resolved). */
+    public ?int $sourceChannelId = null;
 
     public int $seriesPriority = 50;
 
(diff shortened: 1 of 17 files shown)


Grimothy:
Going to be my last contrib for a bit

Serph91P:
> Going to be my last contrib for a bit

Nooo, don't leave us :cry:

Grimothy:
> > Going to be my last contrib for a bit
> 
> Nooo, don't leave us 😢

I was just feeling manic about life that day :D. Not really going anywhere

sparkison:
> Going to be my last contrib for a bit

nooooooo! As long as it's temporary, I'll allow it 😂

PR #1128, opened 2026-05-15
Add duplicate custom playlist

Aldione:


sparkison:
Was missing a couple things, I've went ahead and added directly, so it'll be out in the next release 😉

PR #1133, opened 2026-05-16
feat(sync): Add sync run models to track sync runs and ensure orderin…

sparkison:
# Sync Run Pipeline

## Overview

The sync pipeline replaces the old distributed state machine — where each job decided what ran next by dispatching its own downstream jobs — with a single, ordered `SyncRun` record that defines the full pipeline upfront. A `SyncPipelineService` is the only place that decides "after phase X, dispatch job Y."

This eliminates race conditions, prevents double-firing of jobs, and guarantees correct ordering (e.g. find-replace completes before STRM files are written, so filenames embed corrected titles).

---

### `sync_runs` table

| Column | Type | Notes |
|---|---|---|
| `id` | bigint | Primary key |
| `playlist_id` | bigint | Indexed, cascades on delete |
| `user_id` | bigint | Nullable, indexed |
| `trigger` | varchar(64) | `full_sync` or UI action name |
| `status` | varchar(32) | `pending` / `running` / `completed` / `failed` |
| `phases` | jsonb | Ordered array of `SyncRunPhase` values decided at creation |
| `phase_statuses` | jsonb | Mutable map: `{ "vod_probe": "completed", ... }` |
| `current_phase` | varchar(64) | Nullable |
| `context` | jsonb | `{ playlist_id, user_id }` |
| `progress_message` | text | Nullable, set on failure |
| `started_at` | timestamp | |
| `finished_at` | timestamp | Nullable |

Completed runs older than 30 days are pruned automatically via `MassPrunable`.

---

## Phase Ordering

Phases are built in three ordered groups by `SyncPipelineService::resolvePipeline()`:

**Group 1 — Pre-STRM (metadata + TMDB + probe)**
Runs first for both media types. Populates `stream_stats`, TMDB enrichment, and probe-derived codec/quality data that find-replace rules and STRM filenames depend on.

**Group 2 — FindReplace**
Runs after all metadata and probe phases, before any STRM generation. This ensures `title_custom` is already corrected when STRM filenames are written.

**Group 2b — Live-channel housekeeping (ChannelMerge → LiveProbe)**
`ChannelMerge` resolves failover priorities and duplicates across playlists. `LiveProbe` then dispatches any recurring channel scrubbers (fire-and-forget) and runs `ProbeChannelStreams` against enabled live channels. Both run after FindReplace (so merged channels reflect corrected titles) and before STRM generation.

**Group 3 — STRM generation**
Runs last for both media types. By this point, all enrichment and name corrections are in place.

A full pipeline with all flags enabled (including probe) looks like:

```
Import
VodMetadata → VodTmdb → VodProbe
SeriesMetadata → SeriesTmdb → SeriesProbe
FindReplace
ChannelMerge → LiveProbe
VodStrmPostProbe → SeriesStrmPostProbe
CustomPlaylistSync
SyncCompleted
```

With probe disabled, `VodStrm` / `SeriesStrm` are used instead of their post-probe variants and still run after `FindReplace`.

---

## Phases

Defined in `app/Enums/SyncRunPhase.php`:

| Phase | Value | When included |
|---|---|---|
| `Import` | `import` | Always first when sync is launched via `startImport()` (modern flow) |
| `VodMetadata` | `vod_metadata` | `auto_fetch_vod_metadata` enabled |
| `VodTmdb` | `vod_tmdb` | `tmdb_auto_lookup_on_import` enabled |
| `VodProbe` | `vod_probe` | `auto_probe_vod_streams` enabled |
| `VodStrm` | `vod_strm` | STRM enabled **and** probe **disabled** |
| `VodStrmPostProbe` | `vod_strm_post_probe` | STRM **and** probe both enabled |
| `SeriesMetadata` | `series_metadata` | `auto_fetch_series_metadata` enabled |
| `SeriesTmdb` | `series_tmdb` | `tmdb_auto_lookup_on_import` enabled |
| `SeriesProbe` | `series_probe` | `auto_probe_vod_streams` enabled |
| `SeriesStrm` | `series_strm` | Series STRM enabled **and** probe **disabled** |
| `SeriesStrmPostProbe` | `series_strm_post_probe` | Series STRM **and** probe both enabled |
| `FindReplace` | `find_replace` | Any find/replace or sort-alpha rule enabled |
| `ChannelMerge` | `channel_merge` | `auto_merge_channels_enabled` is true |
| `LiveProbe` | `live_probe` | `auto_probe_streams` is true (live channel probe; bundles recurring scrubbers) |
| `CustomPlaylistSync` | `custom_playlist_sync` | Any auto-sync-to-custom rule enabled |
| `SyncCompleted` | `sync_completed` | Always last |

**Key rule for STRM phases:** STRM sync runs exactly once. When probe is enabled, STRM is deferred until after the probe (`*StrmPostProbe`). This ensures filenames always include probe-derived quality/codec metadata and that find-replace has already corrected `title_custom`.

---

## Service: `SyncPipelineService`

**`startImport(Playlist, trigger)`** *(modern kickoff)*
Called at the very start of a full sync, **before** `ProcessM3uImport` is dispatched. Creates a `SyncRun` with `status = running` and `phases = [Import]` so the UI immediately reflects the sync. The real post-import phases are unknown at this point (channel/series rows don't exist yet) — they are resolved later by `expandPipelineAfterImport()`.

**`expandPipelineAfterImport(SyncRun, Playlist, GeneralSettings)`**
Called from `ProcessM3uImportComplete` once the import chain finishes and the DB is populated. Resolves the real post-import phases and replaces `phases` with `[Import, ...resolved phases..., SyncCompleted]`. The caller then marks `Import` complete via `completePhase()` so the next phase dispatches through normal pipeline progression.

**`buildPipeline(Playlist, GeneralSettings, trigger)`** *(legacy)*
Used only when `ProcessM3uImportComplete` runs with `syncRunId = null` (one-off partial actions that didn't go through `startImport()`). Creates a `SyncRun` from scratch with the resolved phases.

**`startRun(SyncRun)`**
Marks the run as `running` and dispatches the first phase. Used by the legacy path.

**`completePhase(int $syncRunId, SyncRunPhase)`**
Called by gateway jobs when a phase finishes. Wrapped in `DB::transaction()` with `lockForUpdate()` so two workers completing the same phase simultaneously cannot double-dispatch the next one. After the transaction commits, dispatches the next phase or calls `finish()`.

Guards:
- Ignores calls where the run is not `running` (handles retries and stale jobs).
- Ignores phases not in the run's plan.
- Idempotent — no-ops if the phase is already marked `completed`.

**`dispatchPhase(SyncRun, SyncRunPhase)`**
The only place that maps phases to jobs. Uses a `match` expression — adding a new phase requires one line here. `Import` is a no-op because it is driven externally by the `ProcessM3uImport` job chain (not dispatched from the pipeline).

**`chainOrDispatch(array $jobs, SyncRun)`**
Dispatches a single job directly or chains multiple jobs via `Bus::chain()`. Multi-job chains attach a `.catch()` handler that marks the run as `failed` if any job in the chain throws, preventing the run from being left stuck in `running`.

**`finish(SyncRun)`**
Marks the run `completed` and fires the `SyncCompleted` event. Exactly one `SyncCompleted` event fires per full sync.

---

## Job Integration

Every gateway job gains `?int $syncRunId = null`:
- `null` → legacy dispatch path (unchanged, all existing tests remain green)
- non-null → calls `SyncPipelineService::completePhase()` instead of managing downstream dispatch itself

Gateway jobs wired to the pipeline:

| Job | Phase it completes |
|---|---|
| `ProcessVodChannelsComplete` | `VodMetadata` |
| `FetchTmdbIds` | `VodTmdb` or `SeriesTmdb` (via `completionPhase` param) |
| `ProbeVodStreamsComplete` | `VodProbe` or `SeriesProbe` (via `isSeriesProbe` param) |
| `SyncVodStrmFiles` | `VodStrm` or `VodStrmPostProbe` (via `completionPhase` param) |
| `SyncSeriesStrmFiles` | `SeriesStrm` or `SeriesStrmPostProbe` (via `completionPhase` param) |
| `CheckSeriesImportProgress` | `SeriesMetadata` |
| `MergeChannels` (chained with `CompleteSyncPhase`) | `ChannelMerge` |
| `ProbeChannelStreamsComplete` | `LiveProbe` |
| `CompleteSyncPhase` | `FindReplace` or `CustomPlaylistSync` (terminal job in chained phases) |

---

## Import Chain Ordering

A full sync now follows this kickoff lifecycle:

```
[Entry point] → SyncPipelineService::startImport()   ← SyncRun visible immediately, phases=[Import]
             → ProcessM3uImport (carries syncRunId)
             → VOD discovery chunks
             → Series discovery chunks
             → ProcessM3uImportComplete (carries syncRunId)
                 ├─ expandPipelineAfterImport()  ← phases become [Import, ...resolved..., SyncCompleted]
                 └─ completePhase(Import)         ← dispatches first resolved phase
             → ... pipeline progression ...
             → finish()                            ← fires SyncCompleted event
```

All 8 full-sync entry points call `startImport()` first and pass the resulting `syncRunId` into `ProcessM3uImport`:
- `app/Listeners/PlaylistListener.php` (playlist created)
- `app/Console/Commands/RefreshPlaylist.php` (manual + scheduled)
- `app/Console/Commands/ResetSyncProcess.php` (unstick)
- `app/Filament/Resources/Playlists/PlaylistResource.php` (3 refresh actions)
- `app/Http/Controllers/PlaylistController.php` (API refresh + save-resync)

If the import chain fails, `ProcessM3uImport`'s `.catch()` handler calls `SyncPipelineService::fail()` to mark the run as failed (in addition to its existing playlist/notification handling).

`ProcessM3uImportComplete` is the only place that calls `expandPipelineAfterImport()` (modern path) or `buildPipeline()` (legacy path). By the time it runs, both VOD channels and Series rows exist in the DB, so phase resolution correctly reflects what was actually imported.

This guarantees:
- **SyncRun visible from sync kickoff** — UI sees the run the moment a sync starts.
- **One `SyncRun` per full sync** — no mini-pipelines, no overlap.
- **One `SyncCompleted` event per full sync** — fired only from `SyncPipelineService::finish()`.
- **Accurate timeline** — only phases that will actually run are listed; `Import` shows its real elapsed time when complete.
- **Correct FindReplace ordering** — series titles exist before FindReplace runs, so `SeriesStrm` filenames embed corrected names.
- **Concurrent-sync safe** — `syncRunId` is threaded through jobs by ID and never looked up by playlist.

One-off Filament actions (e.g. "Process VOD", "Process Series", manual STRM/TMDB refresh) dispatch the relevant gateway job with `syncRunId = null` and follow the legacy `SyncListener` path — they do not create a `SyncRun`.

---

## `SyncListener` Integration

`SyncListener` skips phases already handled by the pipeline to prevent double-execution:

```php
// Skip find-replace if the pipeline already ran it as a tracked phase
if (! $syncRun?->isPhaseComplete(SyncRunPhase::FindReplace)) {
    $this->dispatchNameProcessingPipeline($playlist);
}

// Skip merge / scrubbers / live-probe pieces the pipeline already ran.
// ChannelMerge covers MergeChannels; LiveProbe covers both scrubbers and ProbeChannelStreams.
$this->dispatchChannelScanJobs(
    $playlist,
    skipMerge: (bool) $syncRun?->isPhaseComplete(SyncRunPhase::ChannelMerge),
    skipLiveProbe: (bool) $syncRun?->isPhaseComplete(SyncRunPhase::LiveProbe),
);

// Skip custom playlist sync if the pipeline already ran it
if (! $syncRun?->isPhaseComplete(SyncRunPhase::CustomPlaylistSync)) {
    $this->dispatchAutoSyncToCustomPlaylistJobs($playlist);
}
```

Post-processes, plugin hooks, Plex DVR sync, and EPG cache regeneration always run from the listener — they are not tracked pipeline phases.

---

## Filament UI

- **`SyncRunResource`** — nested under Playlists, shows all runs for a playlist with status badges, phase timeline, duration, and trigger.
- Navigation to the runs list is via a `view_sync_runs` URL action on the Playlist resource (not a relation manager).
- Phase timeline uses `SyncRunPhase::getLabel()` and `SyncRunPhase::getColor()` for display.
- While a run is `running` and only `Import` is present in `phases`, the timeline renders a synthetic "Post-processing (resolving…)" placeholder row so users see that more steps will follow. The placeholder disappears the moment `expandPipelineAfterImport()` populates the real phase list.


sparkison:
Taking another stab at this one! Trying to get the post processing order in order 😅

PR #1135, opened 2026-05-16
fix: use playlist auths for playlist aliases

djrarky:
Migrates existing playlist alias auth credentials to Playlist Auths, and assigns them to the matching playlist alias.


sparkison:
Gonna close since it's already possible to assign an Auth to an Alias 😉 you can do so from the Auth's management page. I might circle back later and add a select from the Alias page, but I think the current setup works for now and already accomplishes this.

PR #1139, opened 2026-05-19
fix: logo proxy rejecting images when remote omits Content-Type

Warbs816:
## Summary

Some EPG icon hosts return valid image bytes with **no `Content-Type` header** (or a generic `application/octet-stream`). `LogoProxyController::fetchRemoteLogo` rejects those responses because of a strict `str_starts_with($contentType, 'image/')` check, so the proxy ends up serving `placeholder.png` for the affected channels even though the original logo URL is fine.

The fix:

- Move the 5 MB size cap above the type check so oversized responses are still rejected before any parsing.
- When the `Content-Type` header is missing or non-`image/*`, sniff the body with `getimagesizefromstring` and accept the response if it's a recognised image. Small manual SVG fallback included since `getimagesizefromstring` doesn't cover SVG.
- If neither the header nor the body looks like an image, still return `null` so the placeholder is used.

Heads-up: the placeholder response sets `Cache-Control: public, max-age=86400`, so anyone who's been hitting a broken logo will keep seeing the placeholder for up to a day after upgrading until that cache entry expires. Hard-reload doesn't bypass it; incognito does. Happy to follow up with a separate PR shortening that TTL if you'd like.

## Test plan

- [x] On a channel whose EPG icon comes from a host that doesn't set `Content-Type`, the M3U logo and the EPG viewer logo both resolve to the real image instead of the placeholder.
- [x] Channels whose icons come from a host that *does* set `Content-Type: image/png` still work (unchanged behaviour).
- [x] A non-image response (HTML/text) still falls back to the placeholder.

### DIFF
--- a/app/Http/Controllers/LogoProxyController.php
+++ b/app/Http/Controllers/LogoProxyController.php
@@ -124,18 +124,24 @@ class LogoProxyController extends Controller
                 return null;
             }
 
-            $contentType = $response->header('Content-Type');
-
-            // Validate that it's an image
-            if (! str_starts_with($contentType, 'image/')) {
-                return null;
-            }
-
             $content = $response->body();
 
             // Check file size (limit to 5MB)
             if (strlen($content) > 5 * 1024 * 1024) {
                 return null;
+            }
+
+            $contentType = $response->header('Content-Type');
+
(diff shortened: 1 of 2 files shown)


sparkison:
Just added some XSS sanitizing 😉 merging now!

PR #1153, opened 2026-05-27
Fix Xtream auto channel increment numbering

Serph91P:
## Summary
- Make Xtream live stream numbering respect `auto_channel_increment` before imported provider channel numbers.
- Make Xtream VOD stream numbering use the same auto-increment behavior.
- Add regression coverage for live and VOD Xtream outputs with imported provider numbers.

## Testing
- `docker compose -f docker-compose.dev.yml --profile test run --rm --build m3u-editor-dev-test tests/Feature/XtreamAutoChannelIncrementTest.php`
- `docker compose -f docker-compose.dev.yml --profile test run --rm m3u-editor-dev-test tests/Feature/CustomPlaylistSortOrderTest.php tests/Feature/XtreamAutoChannelIncrementTest.php`
- `docker compose -f docker-compose.dev.yml --profile test run --rm --entrypoint /bin/bash m3u-editor-dev-test -lc './vendor/bin/pint app/Http/Controllers/XtreamApiController.php tests/Feature/XtreamAutoChannelIncrementTest.php --dirty'`

### DIFF
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -610,7 +610,9 @@ class XtreamApiController extends Controller
                     $channelNo = ($isCustomPlaylist && ! empty($channel->ccp_channel_number))
                         ? (int) $channel->ccp_channel_number
                         : $channel->channel;
-                    if (! $channelNo && ($playlist->auto_channel_increment || $idChannelBy === PlaylistChannelId::Number)) {
+                    if ($playlist->auto_channel_increment) {
+                        $channelNo = ++$channelNumber;
+                    } elseif (! $channelNo && $idChannelBy === PlaylistChannelId::Number) {
                         $channelNo = ++$channelNumber;
                     }
 
@@ -717,6 +719,7 @@ class XtreamApiController extends Controller
 
             return response()->stream(function () use ($cursor, $playlist, $baseUrl, $isCustomPlaylist) {
                 $num = 0;
+                $channelNumber = $playlist->auto_channel_increment ? $playlist->channel_start - 1 : 0;
                 echo '[';
                 $first = true;
                 foreach ($cursor as $channel) {
@@ -753,6 +756,9 @@ class XtreamApiController extends Controller
                     $vodChannelNo = ($isCustomPlaylist && ! empty($channel->ccp_channel_number))
                         ? (int) $channel->ccp_channel_number
                         : ($channel->channel ?: $num);
+                    if ($playlist->auto_channel_increment) {
+                        $vodChannelNo = ++$channelNumber;
+                    }
 
                     echo json_encode([
                         'num' => $vodChannelNo,
--- a/composer.lock
+++ b/composer.lock
@@ -9856,16 +9856,16 @@
         },
         {
             "name": "symfony/html-sanitizer",
-            "version": "v8.0.13",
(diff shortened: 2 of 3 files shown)


Serph91P:
@sparkison added some security warnings "fixes" aswell because pipline did not complete succesfully.

sparkison:
> @sparkison added some security warnings "fixes" aswell because pipline did not complete succesfully.

Thanks! I'm rolling back those changes and I'll get a package update in as a separate commit.

madasus:
@Serph91P It doesn't look like this ever made it into the DEV version?

Serph91P:
> @Serph91P It doesn't look like this ever made it into the DEV version?

@madasus It did last month as you can this was merged?
https://github.com/m3ue/m3u-editor/commit/3044a0648bdeafac4c36da56cfea0a864eb36b2b

madasus:
@Serph91P i'm running 0.11.58-dev and i'm still seeing the same issue.

This is what i have set which worked when i was using the M3U option

image

Serph91P:
> @Serph91P i'm running 0.11.58-dev and i'm still seeing the same issue.
> 
> This is what i have set which worked when i was using the M3U option
> image

@madasus https://github.com/m3ue/m3u-editor/pull/1215
@sparkison FYI

madasus:
@Serph91P Thanks.  I'm not sure what your reference was above.  Did it not get merged or did it need another fix?

madasus:
@Serph91P any idea what needs to happen here? I just checked the latest DEV version and this is still happening.

sparkison:
@madasus, guessing the channel actually has a number set in the Xtream version of your playlist when it did not in the M3U version. The auto-increment is specifically used when no channel number is set (.e.g; `tvg-chno=""`)

613281935-421fb41d-784a-42d7-8b58-d4e42a7d3668

My guess is that the XAPI is returning a channel number now (`num` key in XAPI) and that's why you're seeing different behavior.

It sounds like maybe you're looking for an override? To output using the auto increment ignoring any provider values?

madasus:
@sparkison possibly - I don't think I have a way to know if they are setting a channel number but that sounds like the issue.  If we can override that would be great.

sparkison:
> @sparkison possibly - I don't think I have a way to know if they are setting a channel number but that sounds like the issue. If we can override that would be great.

Just so I make sure I'm covering this correctly, could I get an example of your output before and after? And were you using Xtream API in your player(s) at all times, even when switching the input types from M3U to XAPI? Or was it different?

Sorry, I'm still not fully understanding how the output has changed so that I can make sure the override works as intended.

madasus:
@sparkison Unfortunately I cannot get back to the before state as the m3u download for this particular service no longer functions.  When I'm using the m3u download with another service i'm seeing the channels picking up the 7000 channel range correctly.  Right now with the Xtream download it is getting random channel numbers all of the place (anywhere from 100 - 4000 i'm seeing).

madasus:
@sparkison anything else I can try to provide on this?

madasus:
@sparkison was the additional logic added to the dev branch?

Serph91P:
> @sparkison was the additional logic added to the dev branch?

@madasus This was merged on may 27. And since has been merged to main as well. So please open an issue or reopen your initial one so we can talk there. The pull request is not the place for this.

PR #1169, opened 2026-06-06
feat: add Filament queue indicator

Serph91P:
## Summary
- Adds a Filament topbar queue indicator that is visible to authenticated users.
- Keeps the Horizon queue manager link restricted to admins when queue manager access is enabled.
- Adds an authenticated JSON endpoint for polling queue state.
- Shows a human-readable overview of queue activity:
  - running job count
  - waiting job count
  - active batches
  - individual running jobs
  - upcoming jobs
  - batch progress, processed jobs, open jobs, failed jobs, and estimated remaining time when available
- Adds optional chunk visibility for chunked jobs, for example `Chunk 5 of 20`, when chunk metadata is available in the queued job payload.
- Moves queue indicator labels and status text to i18n instead of hardcoding German strings.
- Uses English as the primary UI language and adds German translations in `lang/de.json`.
- Updates feature tests for the endpoint, queue snapshot structure, i18n labels, running job counts, batch data, and chunk metadata.

## Implementation notes
- The indicator is registered through the Filament topbar render hook.
- The polling endpoint is protected by normal authentication, so regular logged-in users can understand what background work is happening.
- Access to the full queue manager remains separate from the indicator and is still controlled by admin access plus the existing queue manager setting.
- Chunk metadata is best-effort and optional. Jobs without chunk fields continue to render normally.
- The queue snapshot degrades safely if Horizon or batch data is unavailable.

## Testing
- Rebuilt and restarted the reusable test stack:
  - `docker compose -f compose.test.yaml --env-file .env.test up -d --build m3u-editor`
- Verified `m3u-editor-test` became healthy.
- Verified `/up` returns HTTP 200.
- Verified the route exists:
  - `admin/api/queue-indicator`
  - route name `admin.queue-indicator`
- Verified the queue indicator service can produce a snapshot through Tinker.
- Ran PHP syntax checks for changed PHP and Blade files.
- Validated `lang/en.json` and `lang/de.json` parse correctly.

## Notes
- The local test container does not expose the standard `php artisan test` command in this setup, so full feature tests could not be executed there.
- `php artisan config:cache` in a direct `docker exec` context can fail in the test stack because Reverb/Pusher env values are not injected for that CLI path. App health, `/up`, route lookup, and queue snapshot verification passed.


Serph91P:
@sparkison hope this is an attempt you can agree on :wink:

sparkison:
> @sparkison hope this is an attempt you can agree on 😉

lol! For sure! 😂

however, I went a slightly different route with it, which I think you'll like 😉 ....It'll be in the next release...

PR #1170, opened 2026-06-06
fix: preserve unicode in EPG XML output

Serph91P:
## Summary
- Escape generated EPG XML with explicit UTF-8 handling instead of relying on PHP default_charset.
- Use a shared XML escaping helper for channel, programme, icon, category, episode, and rating fields.
- Add a regression test with Albanian characters and XML-sensitive entities under a non-UTF-8 default charset.

## Testing
- `docker run --rm -v "$PWD":/app -w /app -e APP_ENV=testing -e APP_KEY=base64:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA= -e REVERB_APP_SECRET=test-secret -e DB_CONNECTION=sqlite_testing -e CACHE_STORE=array -e SESSION_DRIVER=array -e QUEUE_CONNECTION=sync composer:2 sh -lc 'touch database/jobs.sqlite .env && trap "rm -f database/jobs.sqlite .env" EXIT && apk add --no-cache icu-dev >/dev/null && docker-php-ext-install bcmath intl pcntl >/dev/null && php artisan test tests/Feature/EpgGenerateControllerTest.php --display-warnings'`
- `docker run --rm -v "$PWD":/app -w /app composer:2 sh -lc 'vendor/bin/pint --test app/Http/Controllers/EpgGenerateController.php tests/Feature/EpgGenerateControllerTest.php'`
- `git diff --check`

### DIFF
--- a/app/Http/Controllers/EpgGenerateController.php
+++ b/app/Http/Controllers/EpgGenerateController.php
@@ -31,6 +31,14 @@ class EpgGenerateController extends Controller
     private const CACHE_TTL_HOURS = 6; // Cache files for 6 hours
 
     /**
+     * Escape XML text and attribute values using UTF-8 explicitly.
+     */
+    private function escapeXml(mixed $value): string
+    {
+        return htmlspecialchars((string) $value, ENT_XML1 | ENT_QUOTES, 'UTF-8');
+    }
+
+    /**
      * Generate the EPG XML file
      *
      * @return Response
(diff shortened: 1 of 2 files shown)


Serph91P:
@sparkison dunno if its actually needed just had someone on the discord post and issue with encoding.

sparkison:
This is a legit issue! Was just missing icons so added that, otherwise looking good! I'll get it merged shortly...

PR #1200, opened 2026-06-19
FetchTmdbIds: prefix-aware VOD matching + chunkById to stop dropped matches

alexdelifer:
## Problem

Two issues cause wrong or missing TMDB matches for VOD on Xtream providers that prefix stream names with category labels (`EN - `, `4K-EN - `, `D+ - `, `NF - `, `EN-TOP - 123. `, …):

1. **Prefixes poison the search.** `searchMovie()` receives the raw prefixed title and `normalizeTitle()` mangles it — e.g. `NF - My Hero Academia` searches as *NFL …*, `4K-EN - The Conjuring` collapses to *En* — yielding confident-but-wrong matches.
2. **Matches silently dropped under SQLite lock.** `processVodChannels()`/`processSeries()` iterate with `cursor()`, which holds an unbuffered read statement open on the same SQLite connection. The per-row `update()` inside the loop then can't upgrade to a writer and throws `database is locked` under any concurrency (queue workers, parallel jobs). The exception is swallowed per-row, leaving a stale/empty `tmdb_id`.

In a ~1800-title library this produced dozens of titles all stuck on one wrong id and ~150 unmatched.

## Changes

- **Prefix-aware lookup:** new `cleanTitleForSearch()` strips the configured VOD name-filter patterns (the same `vod_stream_file_sync_name_filter_patterns` the .strm sync already applies to file names) plus any leading `123. ` numbering, *before* the TMDB search. One setting now cleans both the on-disk name and the match query. Display fields (`title_custom`) are untouched.
- **`cursor()` → `chunkById()`** for VOD and series. Buffered pages close their statement before writes; paging is by primary key, which the updates never change.

## Testing

Built and deployed on a live ~1800-title library. Fresh-import titles that previously mis-matched now resolve correctly (`4K-EN - Tom Clancy's Without Remorse (2021)` → 567189, `NF - My Hero Academia: You're Next` → 1159311, `D+ - Avengers: Endgame` → 299534, `4K-EN - The Conjuring` → 138843). No more dropped-write `database is locked` errors during bulk fetch.

Both changes are behind existing behavior when no name-filter patterns are configured (only the `cursor()`→`chunkById()` swap applies, which is a pure reliability fix).

### DIFF
--- a/app/Filament/Pages/Preferences.php
+++ b/app/Filament/Pages/Preferences.php
@@ -1423,6 +1423,42 @@ class Preferences extends SettingsPage
                                             ->maxValue(100)
                                             ->default(80)
                                             ->helperText(__('Minimum title similarity percentage (50-100) required to accept a match. Higher values = stricter matching.')),
+                                        Fieldset::make(__('Title Cleaning for TMDB Lookup'))
+                                            ->columnSpanFull()
+                                            ->schema([
+                                                Grid::make()
+                                                    ->columnSpanFull()
+                                                    ->columns(2)
+                                                    ->schema([
+                                                        Toggle::make('vod_stream_file_sync_name_filter_enabled')
+                                                            ->label(__('Strip provider prefixes from VOD titles before matching'))
+                                                            ->helperText(__('Remove prefix patterns from VOD titles before searching TMDB (e.g. "EN - ", "4K-EN - ", "NF - ").'))
+                                                            ->live()
+                                                            ->inline(false)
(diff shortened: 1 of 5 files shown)


sparkison:
Thanks for this! Good find! I went ahead and applied to series too since it was missing there, then opened up the options to include the fields (found out they were orphaned through this PR and never used, so I was able to expose them and utilize for matching 😉)

title-cleaning-options

PR #1202, opened 2026-06-19
feat: Sonarr/Radarr download queue monitor + episode interactive search

Grimothy:
## Summary

- **Download queue monitor** with live updates via Reverb WebSocket + 10s polling — shows quality, protocol (NZB/Torrent), indexer, and episode label (S01E05 · Title) per item
- **Precise import states** from `trackedDownloadState`: Import Pending → Importing → Imported; items persist after completion with a dismiss (✕) button; orphan live items are snapshotted so they don't silently vanish
- **Fix: episode request first-attempt failure** — `requestEpisode()` now retries the `/episode` fetch up to 5× (500ms between attempts) after adding a new series, handling Sonarr's async episode indexing
- **Episode-level interactive search** — each episode row has a magnifying glass button that calls `loadEpisodeReleases()`, fetches `/release?seriesId=X&episodeId=Y`, and auto-opens the Interactive Search accordion labelled "Interactive Search — S01E05"; `downloadDetailRelease` passes `episodeId` in the Sonarr payload

## Test plan

- [ ] Request an individual episode from a TV series not yet in Sonarr — should succeed on first attempt (no longer requires a second try)
- [ ] Click the magnifying glass (🔍) next to an episode on a series not in Sonarr — series is added silently, accordion opens with episode-specific releases
- [ ] Click Pick Release on a series already in Sonarr — episode releases load without re-adding
- [ ] Download a release from the episode accordion — verify Sonarr receives `episodeId` in the payload
- [ ] Open queue monitor while a download is active — verify quality, protocol, indexer, and episode label are shown
- [ ] Watch an item complete — it should stay visible with a dismiss button rather than disappearing
- [ ] Dismiss a completed item — it should be removed from the queue view
- [ ] 80/80 ArrSearch tests pass (`php artisan test --filter=ArrSearchTest`)

### DIFF
--- a/app/Events/ArrQueueUpdated.php
+++ b/app/Events/ArrQueueUpdated.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Events;
+
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class ArrQueueUpdated implements ShouldBroadcast
+{
+    use Dispatchable, InteractsWithSockets, SerializesModels;
+
+    public function __construct(public readonly int $userId) {}
+
+    public function broadcastOn(): PrivateChannel
+    {
+        return new PrivateChannel('arr-queue.'.$this->userId);
+    }
+
+    public function broadcastAs(): string
+    {
+        return 'queue.updated';
+    }
+
+    /**
+     * No payload needed — the Livewire component re-fetches on receipt.
+     *
+     * @return array<string, mixed>
+     */
+    public function broadcastWith(): array
+    {
+        return [];
+    }
+}
--- a/app/Filament/GuestPanel/Pages/GuestRequestContent.php
+++ b/app/Filament/GuestPanel/Pages/GuestRequestContent.php
@@ -1,0 +1,131 @@
+<?php
+
+namespace App\Filament\GuestPanel\Pages;
+
+use App\Facades\PlaylistFacade;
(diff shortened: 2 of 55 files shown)


sparkison:
@Grimothy, might need a re-sync with `exp`, the DVR tools are already there, security fix was already pushed earlier, etc.

sparkison:
Think it's in a good spot to merge. I'm not 100% sure about the `usleep()` in `resolveEpisodeId()` method, we might want to move that to a background job or something instead of live-polling Sonarr and blocking.

Also getting a lot of timeout errors when loading details for a season on a fresh cache, this one seems to do it nearly every time 😂

popular

as soon I hit "Select Seasons" (or just click the poster), the panel grays and it takes a while, then I get a 30s timeout hit error. If I refresh and try again it works (guessing because it's cached at this point). Not sure what we can do to provide better UX here though? Didn't get a chance to dig into this specific piece yet though, but guessing it's the TV Maze piece fetching the metadata and maybe I was just hitting the API limit too.

Grimothy:
shameful --- I looked into the agentically   Issue 2 — 30s timeout on detail panel open: The flow on cache miss is lookupTvMazeId (8s timeout) → fetchEpisodes (10s) → fetchCast (10s) —
  sequential, worst case ~28s, which is why it hits Livewire's 30s limit. Fix: once we have the TvMaze ID, parallelize fetchEpisodes + fetchCast
  with Http::pool(). Cache is already 24hr so this only affects first load.

Grimothy:
Fixed in TvMazeService::fetchSeriesData(). The root cause was sequential HTTP calls: lookup →
  episodes → cast, up to ~28s on a cold cache. Now, once the TvMaze ID is resolved, episodes and cast are fetched in parallel via Http::pool(),
  cutting the cold-cache worst case roughly in half (~8s lookup + ~10s for whichever of episodes/cast is slower). If both are already cached the
  pool is skipped entirely. The 24hr cache means most users only hit the cold path once per show.

PR #1203, opened 2026-06-20
fix: plugin schema foreign key delete actions

djrarky:
Apply manifest on_delete options to the generated foreign key constraint rather than the foreignId column definition. This ensures plugin-declared cascade and set-null actions are enforced by the database.

Add a regression test covering cascade behaviour for plugin-declared tables.


sparkison:
Looks good, ty! 🙏

PR #1205, opened 2026-06-20
fix: make scrubber preserve disabled failover channels

djrarky:
Add a protect_failover_channels scrubber setting that defaults on and prevents live checks from re-enabling channels that are registered as failovers.

This lets the scrubber count protected failover channels as live without undoing auto-merge failover deactivation. Add focused coverage for the protected and opt-out behaviours.


djrarky:
Handled in #1206

PR #1210, opened 2026-06-22
feat: plugin run result exports after table prs

djrarky:
## Depends On

- #1208
- #1209

## Simple Explanation

This PR improves host support for plugin-declared result tables created by plugin runs.

It does three related things:

- Shows plugin-declared result tables on a plugin run detail page.
- Scopes those result rows to the selected run and playlist when possible.
- Lets plugins choose which on-demand export formats are available for each declared table.

This is useful for plugins that create many native audit or result rows during a run. For example, a sync plugin may produce rows for matched channels, excluded channels, skipped items, warnings, or per-channel decisions.

In those cases, a global Data tab table is hard to scan because users usually want to inspect one run at a time. The run detail page is the natural place to review those rows.

Exports are generated when the user asks for them. The host does not store CSV or JSON files, so there is no export storage, rotation, or cleanup lifecycle to manage.

## Why This Is Required

Plugin manifests can already declare native tables through `schema.ui_tables`.

Some of those tables are regular plugin configuration tables. Others are more naturally run result tables. A run result table is usually connected to:

- A specific plugin run.
- A specific playlist.
- A user-facing review workflow.
- A potentially large number of result rows.

For example, a plugin may need:

- A row for every channel evaluated during a playlist sync.
- A decision such as matched, excluded, skipped, or changed.
- A score or reason for each decision.
- The rows to be reviewed for one run without mixing in older runs.
- The rows to be downloadable for offline inspection or support/debugging.

Previously, plugin-declared table rows were easiest to view through the plugin Data tab. That works for smaller configuration tables, but it becomes awkward for run output because all runs share the same table UI. Users then have to filter or page through a large table without a clear run-level context.

This PR makes plugin-declared result tables available from the plugin run detail page when the table includes `extension_plugin_run_id`. If the run payload contains `playlist_id`, the table is also scoped to that playlist.

Plugins can declare a run result table like:

```json
{
  "id": "run_results",
  "label": "Run Results",
  "description": "Rows captured by plugin runs.",
  "table": "plugin_example_run_results",
  "export_formats": ["csv"],
  "create": false,
  "edit": false,
  "delete": false,
  "columns": [
    {
      "name": "created_at",
      "label": "Created",
      "type": "datetime",
      "sortable": true
    },
    {
      "name": "result_type",
      "label": "Type",
      "sortable": true
    },
    {
      "name": "decision",
      "label": "Decision",
      "sortable": true
    },
    {
      "name": "title",
      "label": "Title",
      "searchable": true
    },
    {
      "name": "score",
      "label": "Score",
      "sortable": true
    }
  ],
  "fields": []
}
```

When the table has an `extension_plugin_run_id` column, the host can discover it as a run-scoped result table and render it under the related plugin run.

The table remains plugin-declared. The host only provides the run context, read-only presentation, filtering, and export endpoints.

Plugins can control exports per `schema.ui_tables` entry with `export_formats`:

- `["csv"]` enables CSV only.
- `["json"]` enables JSON only.
- `["csv", "json"]` enables both formats.
- `[]` disables export buttons and download routes for that table.

If `export_formats` is omitted, plugin-declared tables keep the existing default of CSV and JSON for compatibility.

## Changes

- Discovers plugin-declared tables that include `extension_plugin_run_id`.
- Shows run-scoped plugin result tables on plugin run detail pages.
- Scopes result table rows by plugin and run.
- Scopes result table rows by playlist when the run payload includes `playlist_id`.
- Keeps run detail result tables read-only.
- Reuses plugin-declared table labels and descriptions.
- Reuses plugin-declared table columns for the run detail table.
- Adds plugin-controlled `export_formats` support for declared table exports.
- Defaults omitted `export_formats` to CSV and JSON for compatibility.
- Allows plugins to disable declared table exports with `export_formats: []`.
- Generates exports on demand instead of storing generated files.
- Adds table filters for run and playlist when those scopes are not already fixed.
- Adds table filters for `result_type` and `decision` when those columns exist.
- Adds a dedicated controller for plugin table export downloads.
- Adds routes for plugin table export downloads.
- Adds regression coverage for run-scoped result table rendering.
- Adds regression coverage for CSV export output.
- Adds regression coverage for JSON export output.
- Adds regression coverage for per-table export format gating.
- Adds regression coverage for disabled per-table exports.
- Adds regression coverage for plugin run detail page result table discovery.

## Testing

Focused tests run on the stacked PR branch:

- `PluginDeclaredTableUiTest`

Result:

- 19 passed, 78 assertions


sparkison:
Alright @djrarky, that wraps it! I've gotten the complete set of PR's merged 😉 this will roll out with the next release 🎉

djrarky:
Mate you are a legend, thanks!

There is one thing tho, pretty much always I wish to export the entirety, not just what's visible in the UI, is there a reasonable way to add that?

sparkison:
> Mate you are a legend, thanks!
> 
> There is one thing tho, pretty much always I wish to export the entirety, not just what's visible in the UI, is there a reasonable way to add that?

aw shit! Saw this too late and missed the release. I'll update for the next one to include all columns.

PR #1212, opened 2026-06-24
feat: clickable cast filmography + automatic search release check

Grimothy:
# feat: clickable cast filmography + automatic search release check

> **Branch:** `Grimothy:Media-Request-Integration` → `m3ue:experimental`
> **Stats:** 47 files changed, +2,929 / −728
> **Commits:** 14 ahead of upstream/experimental
> **Reviewers:** waiting on author to mark ready

---

## Overview

Three cohesive features on top of `upstream/experimental`, each independently testable:

1. **Clickable Cast Members → Actor Filmography** — guests/admins can drill into an actor's full TMDB combined-credits from any detail panel, then jump straight into requesting a specific title.
2. **» No Approved Releases « — MonitorArrSearch job** — automatic-search flows that find nothing usable now actively warn the user instead of silently failing.
3. **Guest Content Request Workflow with Admin Approval Gate** — guests can request Sonarr/Radarr content on shared playlists; admins approve/reject from the existing Arr queue monitor.

Plus two reliability fixes that came out of testing the new flows:
- **Discover → Request** now searches *all* configured integrations (not just the first) and falls back to a title search for newer/foreign titles.
- **Arr form** no longer overwrites an already-set quality profile / root folder when "Test Connection & Discover" is clicked, and the profile select auto-fills the human-readable name on change.

---

## 1. Clickable Cast Members → Actor Filmography

Cast members in the ArrSearch detail panel are now clickable. Clicking an actor opens a dedicated filmography page showing their full TMDB combined-credits (movies + TV, sorted by date). Clicking any filmography item auto-navigates to the request-content page, auto-searches, and opens the detail side-sheet.

### Files

**Service layer**
- `TmdbService`: added `getPersonDetails()`, `getPersonCombinedCredits()`, `searchPersonIdByName()`, `getTvCast()` (TVMaze cast fallback), added `id` field to `getMovieCast()` cast map
- `TvMazeService`: added `id` field to `parseCast()` cast map; bumped cache key to `v2` to bust stale entries without personId

**Filament pages**
- `app/Filament/Pages/ActorFilmography.php` (admin) — extends `Page`, accepts `?personId&name=`, with name-based TMDB lookup fallback
- `app/Filament/GuestPanel/Pages/GuestActorFilmography.php` (guest) — same flow, hidden from nav (`shouldRegisterNavigation = false`)
- `resources/views/filament/pages/actor-filmography.blade.php`
- `resources/views/filament/guest-panel/pages/actor-filmography.blade.php`

[...]

PR #1216, opened 2026-06-26
🐛 fix(dvr): guard retention pruning + warn on empty EPG scope, refactor BrowseShows channel-options dispatch

Grimothy:
## DVR fixes + BrowseShows channel-options refactor

Two small but high-impact changes on top of `upstream/experimental`:

### 1. DVR retention pruning — guard against recordings without a `file_path` (`3035046e`)

`DvrRetentionService::trimGroup()` and `enforceRetentionDays()` now skip recordings where `file_path IS NULL`. Prevents the retention logic from issuing delete commands for in-progress / failed recordings that were never actually captured to disk.

### 2. DVR scheduler — warn when series rule has no EPG channel scope (`3035046e`)

`DvrSchedulerService::matchSeriesRule()` now logs a `warning` (with rule_id, series_title, epg_channel_id, channel_id, source_channel_id) instead of silently returning when the resolved EPG channel scope is empty. The rule had been silently dropped; this surfaces the misconfiguration.

### 3. BrowseShows channel options — dispatch via browser event (`39687530`)

**Before:** `$channelOptions` was a public property eagerly populated on mount and serialized into the initial page HTML payload. For DVR settings with thousands of channels, this bloated page size.

**After:** `loadChannelOptions()` now dispatches a `channel-options-loaded` browser event with the options map. The Alpine.js `x-data` blocks in the three browse-shows Blade templates read from `window['__channelOptions_<dvr_setting_id>']`, populated via `@channel-options-loaded.window="..."`. The flag `$channelOptionsDispatched` prevents redundant queries on repeat dropdown opens.

Two follow-on changes were needed:
- `BrowseShowsTest.php`: 2 tests updated to assert the dispatched event payload via `assertDispatched('channel-options-loaded', fn(...) => ...)` instead of reading the removed property.
- `browse-shows.blade.php`: removed the unused `'channelOptions' => $this->channelOptions` include arg passed to the `browse-show-detail` partial (the detail view now reads from the window event, so the variable was dead and triggered `PropertyNotFoundException` during test renders).

---

### Files changed

| File | Δ |
|---|---|
| `app/Filament/Pages/BrowseShows.php` | +46 / −11 |
| `app/Services/DvrRetentionService.php` | +2 / −1 |
| `app/Services/DvrSchedulerService.php` | +8 / 0 |
| `resources/views/filament/guest-panel/pages/browse-shows.blade.php` | +13 / −12 |
| `resources/views/filament/pages/browse-show-detail.blade.php` | +52 / −16 |
| `resources/views/filament/pages/browse-shows.blade.php` | +11 / −6 |
| `tests/Feature/BrowseShowsTest.php` | +13 / −8 |

**Net:** 7 files, +142 / −57.

### Commits

| Hash | Type | Description |
|---|---|---|
| `3035046e` | 🐛 fix | guard retention pruning and warn when series rule has no EPG scope |
| `39687530` | ♻️ refactor | dispatch channel options via browser event instead of serializing into HTML |
| `944e7d4b` | merge | sync upstream/experimental into dvr-fixes |

### Test coverage

| Suite | Status |
|---|---|
| `DvrRetentionServiceTest` | 28 passed |
| `DvrSchedulerServiceTest` | 30 passed |
| `BrowseShowsTest` | 21 passed |
| `BrowseShowsIsNewTest` | 16 passed |
| **Total affected** | **95 passed (151 assertions)** |
| Pint | clean on all 7 changed files |

### Notes for reviewers

- The merge commit `944e7d4b` is a `git merge upstream/experimental --no-edit` to keep this branch in sync with `m3ue:experimental` after the merge of #1212 (which landed upstream). No code conflicts; the upstream changes touched disjoint files from this branch's work.
- The BrowseShows refactor is a pure performance/scalability improvement — no behaviour change for the happy path.
- DVR retention guards are defensive: they only filter out rows where `file_path IS NULL`, which is an edge case (recording was scheduled/captured but file cleanup removed the path). Existing recordings with valid paths are unaffected.


sparkison:
Got one more fix to push, then we should be good! I'm moving the entire form component to a Filament form/schema so we can use the same schema used in all other forms. Should make it much more performant, and a lot easier to troubleshoot going forward 😉

Grimothy:
This is the exact type of polish I need you brother :D

Grimothy:
also @sparkison  can you add some .md agent rules to gate (to the best of our ability) the particular forms and styles

sparkison:
> also @sparkison can you add some .md agent rules to gate (to the best of our ability) the particular forms and styles

For sure! I'll see if I can add some rules for this...

PR #1223, opened 2026-06-27
fix: add active live group recount bulk action

Serph91P:
## Summary
- Add a separate Live Groups bulk action, `Recount Active Channels in Selected Groups`.
- Recount selected live groups by `groups.sort_order`, then name, then id.
- Number only enabled live channels in each group by `channels.sort`; disabled channels keep their existing channel numbers.
- Keep the existing `Recount Selected Groups` action unchanged.

## Testing
- `python3 -m json.tool lang/en.json lang/de.json lang/es.json lang/fr.json lang/zh_CN.json`
- `git diff --check`
- Docker test container PHP lint for changed PHP/test files
- `vendor/bin/pint --test tests/Feature/GroupEnableChannelRecountTest.php`
- `vendor/bin/pint --test`
- `tests/Feature/GroupEnableChannelRecountTest.php` in Docker test profile: 8 warnings, 24 assertions, exit code 0
- Full Docker Pest suite with embedded Redis: 1523 passed, 29 skipped, 4019 assertions
- `composer audit --locked --format=summary`: no security vulnerability advisories found
- `npm audit --audit-level=high`: 0 vulnerabilities

## Local build note
- Retried local `docker build` twice. Both attempts were blocked before application build by transient GitHub codeload HTTP/2 400 errors while Composer downloaded dist archives, first `igorw/evenement`, then `symfony/polyfill-php84` / `php-fig/log`. The Docker test image build, frontend build inside that image, focused tests, full suite, Pint, and audits passed.

### DIFF
--- a/app/Facades/SortFacade.php
+++ b/app/Facades/SortFacade.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static null bulkSortGroupChannels(\App\Models\Group $record, string $order = 'ASC', ?string $column = null)
  * @method static null bulkRecountGroupChannels(\App\Models\Group $record, int $start = 1)
  * @method static null bulkRecountGroupsByOrder(\Illuminate\Database\Eloquent\Collection $groups, int $start = 1)
+ * @method static null bulkRecountActiveLiveGroupsByOrder(\Illuminate\Database\Eloquent\Collection $groups, int $start = 1)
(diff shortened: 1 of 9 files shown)


sparkison:
Feels like this should be applied to the individual action as well as the VOD groups since they share the same actions. I'm going to close this in favor of enhancing the existing recount actions with a toggle to filter by enabled only so we can merge this feature into the existing recount actions without needing to add more 😉

PR #1253, opened 2026-07-08
media-network-fixes: audio-language/subtitle toggles, HLS controller, Emby seek, proxy auto-transition

Grimothy:
## Summary
Media-network fixes and admin UI improvements: HealStaleQueueMonitors, audio-language/subtitle toggles, network content relations UI, single-authority Emby seek, proxy auto-transition.

## Commits
- \`436ce87d\` feat(admin): HealStaleQueueMonitors command + CopilotTools + audio-language/subtitle toggles
- \`f4dd608a\` feat(networks): admin UI for content relations, schedule regeneration, HLS controller
- \`2988f57d\` feat(content): add pin fields and chain_id for network content relations
- \`db98cd37\` fix(media-server): single-authority seek + byte-range fast-seek for Emby broadcasts
- \`bea748a3\` fix: add eager loading for contentable in EPG/Xtream API queries and fix empty logo fallback
- \`606daf35\` fix(network-broadcast): patch three bugs in proxy auto-transition path
- \`c48f2f46\` Add proxy auto-transition for zero-round-trip programme switching

## Notes
- Targets \`experimental\` per project PR convention.
- Companion to m3u-proxy audio stream + subtitle drift work.
- No destructive migrations (pin/chain_id fields are additive).


sparkison:
Ok, I'm really struggling to review this one, but there's so much going in so many different directions that I can't review comfortably without worrying I'm gonna break things 😅

I did find a couple things that I think could help improve/simplify this though...

The core audio/subtitle feature can be much simpler (I think, we'll need to test to confirm). FFmpeg's `-map 0:a:m:language:fra?` metadata filter selects audio by language natively in Local/Direct mode so no Emby MediaStreams lookup, no remux URL detection, no byte-offset calculation needed. I'd suggest starting there. The current approach builds complexity by resolving the stream index in Laravel, which forces Emby into remux mode, which then breaks seeking and requires the byte-offset workaround. It's worth avoiding that chain if we can!

Recommended split:

1. **Audio + subtitle basic toggles** — two migrations, two UI fields, pass `preferred_audio_language` string + `subtitles_enabled` bool to the proxy and let FFmpeg/proxy handle the rest. This delivers the stated goal.
2. **Emby MediaStreams integration** (separate PR, optional) — if there's a specific need to handle Server transcode mode audio selection or external sidecar subtitle files, that's a separate feature with its own tradeoffs worth reviewing on its own.
3. **Proxy auto-transition** / `next_stream_config` — separate optimization PR.
4. **Content chaining + pinning** — separate feature PR, these are unrelated to audio/subtitle selection.
5. **HealStaleQueueMonitors + Copilot tools** — housekeeping/tooling PR.

The EPG key collision fix (`$network->id` as array key instead of `channel_number` in `EpgApiController`) is a genuine standalone bug fix, I'll go ahead and get this implemented directly.

Grimothy:
fair enough 🫠. I admit i got lost in the sauce

PR #1255, opened 2026-07-08
Fix manual TMDB search result click outside List pages (extract AppliesTmdbSelection trait)

### LINKED ISSUE #1254: Manual TMDB search: clicking a result fails with MethodNotFoundException outside the List pages
### Environment

- Image: `sparkison/m3u-editor:experimental-0.12.36` (also present on 0.12.35 and current `experimental` HEAD)

### Description

The **Manual TMDB Search** action opens fine from anywhere it is exposed, and searching works — but *clicking a search result to apply it* only works when the modal is hosted by the VOD/Series **List** pages. From any other hosting component the click fails (the spinner just keeps loading) and the log shows:

```
production.ERROR: Unable to call component method. Public method [applyTmdbSelection] not found on component
Livewire\Exceptions\MethodNotFoundException at vendor/livewire/livewire/src/Mechanisms/HandleComponents/HandleComponents.php:693
```

### Cause

The result buttons in `resources/views/forms/components/tmdb-search-results.blade.php` do:

```blade
wire:click="applyTmdbSelection({{ $result['id'] }}, '{{ $type }}', {{ $recordId ?? 'null' }}, '{{ $recordType }}')"
```

i.e. they call `applyTmdbSelection` on **whatever Livewire component hosts the modal**. That method (plus its helpers `applyVodMetadata` / `applySeriesMetadata`) is defined only on `ListVod` and `ListSeries` — as two byte-identical ~350-line copies.

But the `manual_tmdb_search` action is reachable from more components than the two List pages, since `VodResource::getTableActions()` / `SeriesResource::getTableActions()` are reused elsewhere. Affected hosts without the handler:

- `ViewVod` / `ViewSeries` (where we hit it in production)
- `CustomPlaylists\RelationManagers\VodRelationManager` and `SeriesRelationManager`
- `MediaServerIntegrations\RelationManagers\MoviesRelationManager` and `SeriesRelationManager`

### Suggested fix

Extract the three (identical) methods into a shared trait and `use` it in every component that can mount the action. That fixes all six broken hosts and also de-duplicates the two existing List-page copies (~700 duplicated lines → one trait).

PR incoming. We have been running the same handler code mounted onto the View pages in production since 2026-07-06 — result clicks apply metadata correctly there and the exception is gone from the logs.

tjelite1986:
Fixes #1254

### Problem

The `tmdb-search-results` component calls `applyTmdbSelection(...)` on its hosting Livewire component, but the handler (and its `applyVodMetadata` / `applySeriesMetadata` helpers) only exists on `ListVod` and `ListSeries`. Since `getTableActions()` is reused by the View pages' hosting contexts and by four relation managers, applying a manual TMDB match from any of those throws:

```
Livewire\Exceptions\MethodNotFoundException: Unable to call component method.
Public method [applyTmdbSelection] not found on component
```

### Change

- The two existing copies on `ListVod` / `ListSeries` were **byte-identical**, so they are extracted verbatim into a new `App\Traits\AppliesTmdbSelection` trait (no logic changes).
- The trait is used by every component that can host the `manual_tmdb_search` modal:
  - `ListVod`, `ListSeries` (unchanged behavior, now via trait)
  - `ViewVod`, `ViewSeries`
  - `CustomPlaylists\RelationManagers\VodRelationManager`, `SeriesRelationManager`
  - `MediaServerIntegrations\RelationManagers\MoviesRelationManager`, `SeriesRelationManager`
- Removed the imports on the List pages that became unused after the move.

Net effect: −317 lines (~700 duplicated lines collapse into one 378-line trait) and the modal works from all eight hosts.

### Testing

- `php -l` clean on all nine touched files (PHP 8.4).
- We have been running these exact handler methods mounted onto `ViewVod` / `ViewSeries` in production (arm64, 0.12.36) since 2026-07-06: result clicks apply full metadata (IDs, poster, plot, genre/group, runtime, etc.) and the `MethodNotFoundException` no longer appears in the logs.


sparkison:
Another nice find! Thanks for the fix 🤘

PR #1262, opened 2026-07-10
fix(alias): swap provider credentials for M3U-imported playlists

### LINKED ISSUE #1261: Playlist aliases - provider credentials for M3U
## Describe the bug

Aliases for m3u file imported playlists don't stream on the client. I think this is because I'm trying to use the provider credentials feature of the alias, but that only supports xtream playlists? I

## To Reproduce

1. Import an M3U playlist.
2. Create an alias for that playlist.
3. Use provider credentials to change credentials used for playlist.
4. This results in streams not loading on the client.

## Expected behavior

Provider credentials should replace the aliased playlist's credentials whether it is xtream or an imported m3u.

## Actual behavior

Provider credentials dont seem to be replaced. I think whichever credentials are in the file are what gets used.

tjelite1986:
## Summary

Fixes #1261

Playlist alias provider credentials were only applied to Xtream API playlists. For M3U-imported playlists, `PlaylistAlias::transformChannelUrl()` / `transformEpisodeUrl()` returned early because the source playlist has no stored `xtream_config`, so the original URLs (with the original credentials) were streamed and the alias credentials were silently ignored.

## Changes

- When the effective playlist has no `xtream_config`, fall back to parsing the base URL and credentials directly from the stream URL itself (M3U exports from Xtream providers keep the standard URL shape).
- `transformUrl()` now also handles the prefix-less live form (`http://host:port/username/password/<id>`) that Xtream `get.php` M3U output commonly uses, in addition to the existing `/(live|series|movie)/` form.
- The prefix-less parser requires a numeric stream ID (optionally with an extension), so arbitrary three-segment URLs (e.g. `https://cdn.example.com/hls/stream/playlist.m3u8`) are left untouched.
- Xtream playlists are unaffected: the stored source config still takes precedence, and `findXtreamConfigByUrl()` matching works the same for parsed configs.

## Tests

New `tests/Unit/PlaylistAliasM3uCredentialSwapTest.php` covering:
- credential swap for M3U playlist channels with prefixed Xtream URLs
- credential swap for the prefix-less live form
- non-Xtream M3U URLs left unchanged
- episode URL swap for M3U playlists
- existing Xtream-config-based swap still working

`php artisan test --testsuite=Unit` passes (149 tests), alias feature tests pass (54 tests), Pint clean.

### DIFF
--- a/app/Models/PlaylistAlias.php
+++ b/app/Models/PlaylistAlias.php
@@ -552,7 +552,12 @@ class PlaylistAlias extends Model
     }
 
     /**
-     * Transform channel URL to use this alias's provider config
+     * Transform channel URL to use this alias's provider config.
+     *
+     * For M3U-imported playlists that have no stored xtream_config, credentials are
+     * parsed directly from the stream URL. The extracted provider URL must match one
+     * of the alias's configured providers — this is both the selection mechanism for
+     * multi-provider aliases and the guard against rewriting non-Xtream CDN URLs.
      */
     public function transformChannelUrl(Channel $channel): string
     {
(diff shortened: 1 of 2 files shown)


sparkison:
great addition! Looks good, I'll try and get this merged for the next release...

PR #1268, opened 2026-07-11
feat(content): pin fields and chain_id for network content relations

Grimothy:
## Summary
Split out of #1253 per review feedback (item 4 in the requested split).

- Add `pin_day_of_week`/`pin_time_of_day` and `chain_id` columns to `network_content`, for weekly time-slot pinning and consecutive playback chaining.
- `NetworkScheduleService` pre-places pinned content at anchored day+time slots and groups chained items so they always play as one shuffle unit; also picks up VOD title/description/poster from the richer metadata fields where available.
- `NetworkContentRelationManager`: inline weight editing, chain_id indicator column, pin day/time columns, and Link/Unlink Chain bulk actions.
- `NetworkResource`: "Generate Schedule" now offers continue-from-current-position vs. fresh-start modes; `media_server_integration_id` locks once content has been added to a network.
- `NetworkContentPinTool`: Copilot tool to pin/unpin content to a weekly timeslot by `network_content_id`, registered in Preferences and documented in the Copilot system prompt.

**Deviation from the suggested split:** `NetworkContentPinTool` was moved here from the housekeeping/CopilotTools PR (#1267), since it hard-depends on the `pin_day_of_week`/`pin_time_of_day` columns added in this PR and can't be tested independently of them.

## Test plan
- [x] Built independently off `experimental`
- [x] `vendor/bin/pint` clean
- [x] Own test suite passes: `NetworkContentChainRelationManagerTest`, `NetworkContentChainTest`, `NetworkContentPinSchedulingTest`, `NetworkContentScheduleRegenerationTest`, `NetworkContentWeightInlineEditTest` (34 passed)
- [x] Existing `NetworkResourceBulkBroadcastActionsTest` still passes (no regression from the Generate Schedule action changes)

### DIFF
--- a/app/Filament/CopilotTools/NetworkContentPinTool.php
+++ b/app/Filament/CopilotTools/NetworkContentPinTool.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Filament\CopilotTools;
+
+use App\Models\Network;
+use App\Models\NetworkContent;
+use EslamRedaDiv\FilamentCopilot\Tools\BaseTool;
+use Illuminate\Contracts\JsonSchema\JsonSchema;
+use Laravel\Ai\Tools\Request;
+use Stringable;
+
+/**
+ * Copilot tool for pinning network content to a specific day and time.
+ *
+ * Allows the AI to set or clear weekly time pins on content items
+ * in a network playlist (e.g. "play The Wild Robot every Friday at 8pm").
+ */
+class NetworkContentPinTool extends BaseTool
+{
+    public function description(): Stringable|string
+    {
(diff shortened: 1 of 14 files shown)


Grimothy:
Added a fixup commit (`c998e1e9`) to address a crash discovered while validating this PR:

**Bug**: `NetworkScheduleService::generateSchedule()` crashed with an undefined-array-key error when a network contained ONLY pinned content (no rotation items). The early-return guard at line 60 only triggers when both rotation content AND pinned content are empty, so the all-pinned case slipped through and the rotation loop ran against an empty `$contentItems`.

**Fix**: Added a 9-line guard inside the transaction closure that exits early when `$contentCount === 0`. Pinned occurrences are pre-placed above the rotation loop, so the timestamp update + programme count return remain correct.

**Tests**: 2 new regression tests in `NetworkContentPinSchedulingTest` covering both `sequential` and `shuffle` schedule types. Full PR test suite passes (39/39).

PR #1269, opened 2026-07-11
fix: apply auto-sort channel numbers to existing channels on sync

### LINKED ISSUE #1264: Channel auto sort from provider does not add a sort number.
## Describe the bug

As title states, when channel sort from provider is enabled, it does not add a sort number, thus breaking in some players that depend on it.

## To Reproduce

1. Enable auto channel sort from provider.
2. Save changes 
3. Sync playlist
4. See no sort number is attached.

## Expected behavior

A sort number to be added.

## Actual behavior

No sort number is added

## Screenshots

N/A

## docker-compose.yml contents

```yaml
# docker-compose.yml contents here
```

## Container/Host Logs

```log
<!-- Insert logs here -->
```

tjelite1986:
## Summary

Fixes #1264 — enabling **Auto channel sort** on an already-imported playlist never attached a sort number.

**Root cause:** `ProcessM3uImportChunk` / `ProcessM3uVodImportChunk` unconditionally exclude `sort` from the upsert update column list. New channels get their sort on insert, but existing channels are never updated — so toggling auto-sort on an existing playlist is a no-op, and provider order changes are never propagated on re-sync (despite the sort-column tooltip saying *"Playlist auto-sort enabled; any changes will be overwritten on next sync"*).

**Fix:** payload rows carry a `sort` value exactly when the import ran with auto-sort enabled (`$channelFields['sort']` is only set then), so the chunk jobs now append `sort` to the update list when the rows have the key. When auto-sort is disabled the rows have no `sort` key and manual sort orders remain untouched, preserving the existing user-override behavior.

## Tests

New `ProcessM3uImportChunkSortTest` (3 cases):
- existing channel gets the payload's sort value when rows carry one (live chunk)
- existing channel's manual sort is untouched when rows carry none
- same propagation through the VOD chunk

`ProcessM3uImportReimportTest`, `XtreamAutoChannelIncrementTest`, `ProcessM3uImportEnableChannelsTest`, `ProcessM3uImportCompleteVodGuardTest` all pass (12 tests), Pint clean.

### DIFF
--- a/app/Jobs/ProcessM3uImport.php
+++ b/app/Jobs/ProcessM3uImport.php
@@ -1183,6 +1183,7 @@ class ProcessM3uImport implements ShouldQueue
         if ($playlist->auto_sort_groups) {
             $groupOrder = 1;
         }
+        $autoSort = $playlist->auto_sort;
 
         // Determine if we should create the channels and groups in the database
         $preProcessingLive = $this->preprocess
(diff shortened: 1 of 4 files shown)


sparkison:
Confirmed! Thanks for the fix 😉

tjelite1986:
> Confirmed! Thanks for the fix 😉

Will ask claude to help fix more issue :)

sparkison:
> > Confirmed! Thanks for the fix 😉
> 
> Will ask claude to help fix more issue :)

Lol, sounds good 😂

PR #1274, opened 2026-07-11
feat: auto enable/disable channels by title/name pattern on sync

### LINKED ISSUE #1199: Enable/Disable Channels Based On Event/Name
**Option 1:**

Ability to enable/disable streams based on activity or event times. 

Example: When a game or event is on, the channel is enabled, but when the event is not on, the channel is disabled. 

This can be accomplished through pattern analysis of channel names or EPG data.  

**Option 2:**

Enabling/disabling channels based on channel title. This can be accomplished through regex filtering.

Example when excluding "NO EVENT" -
LIVE EVENT 01 | EVENT TITLE - Enabled
LIVE EVENT 02 | NO EVENT TODAY - Disabled

These features would allow the user to have a "filtered" channel guide, as long as their provider is feeding correct information and the regex expression is correct.

tjelite1986:
## Summary

Implements option 2 of #1199: per-playlist rules that automatically enable/disable channels based on a regex match against the channel title or name, re-evaluated after **every** sync — so event channels that providers rename between syncs (`LIVE EVENT 01 | EVENT TITLE` ⇄ `LIVE EVENT 01 | NO EVENT TODAY`) flip automatically without manual curation.

## Semantics

- Rules run in order; the **last matching rule wins** per channel.
- Channels matching **no rule keep their current state** (manual enables/disables are untouched).
- This composes naturally: a broad `disable` rule followed by a narrower `enable` rule gives "only enable channels currently carrying an event".
- Matches against the effective value (`title_custom ?? title` / `name_custom ?? name`), so it plays nicely with Find & Replace output.

## Implementation

- `channel_enable_rules` JSON column on `playlists` + repeater UI in the playlist form (new *Auto Enable/Disable Rules* section next to Find & Replace), with target (Live/VOD), column (title / tvg-name), action (enable/disable) and pattern with the existing `RegexTesterAction`.
- New `RunPlaylistChannelEnableRules` job: chunked over the playlist's non-custom channels, bulk-updates only changed rows, skips (and reports) invalid regex patterns without aborting the rest, and triggers `SyncPlexDvrJob::dispatchIfConfigured` when anything changed (mirroring the bulk enable/disable actions).
- New `ChannelEnableRules` sync phase, included only when an enabled rule exists — placed **after FindReplace** (rules match cleaned titles) and **before ChannelMerge/LiveProbe** (those phases see the final enabled set).

Option 1 from the issue (EPG/event-time-based enabling) is intentionally out of scope here, but this phase would be the natural place to extend if that's wanted later.

## Test plan

- New `tests/Feature/ChannelEnableRulesTest.php` (10 tests): disable/enable matching, last-match-wins ordering, custom-title and name-column matching, invalid-regex resilience, live/VOD target isolation, custom channels untouched, and pipeline inclusion/ordering + omission when no rules are enabled.
- `php artisan test --compact --filter='ChannelEnableRulesTest|SyncPipelineServiceTest|SyncListenerTest|ProcessM3uImportCompleteVodGuardTest'` → 99 passed.
- `vendor/bin/pint --dirty` clean.

Closes #1199

### DIFF
--- a/app/Enums/SyncRunPhase.php
+++ b/app/Enums/SyncRunPhase.php
@@ -19,6 +19,7 @@ enum SyncRunPhase: string
     case SeriesStrmPostProbe = 'series_strm_post_probe';
 
     case FindReplace = 'find_replace';
+    case ChannelEnableRules = 'channel_enable_rules';
     case ChannelMerge = 'channel_merge';
     case LiveProbe = 'live_probe';
     case CustomPlaylistSync = 'custom_playlist_sync';
@@ -40,6 +41,7 @@ enum SyncRunPhase: string
             self::SeriesProbe => 'Series Stream Probe',
             self::SeriesStrmPostProbe => 'Series STRM Files (Post-Probe)',
             self::FindReplace => 'Find & Replace / Sort',
+            self::ChannelEnableRules => 'Channel Enable/Disable Rules',
             self::ChannelMerge => 'Channel Merge',
             self::LiveProbe => 'Live Stream Probe',
             self::CustomPlaylistSync => 'Custom Playlist Sync',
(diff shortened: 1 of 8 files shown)


sparkison (commented):
Looks good!

PR #1279, opened 2026-07-12
Fix numeric dynamic select payload validation

Serph91P:
## Summary

- accept integer payload values for dynamic plugin selects while preserving string values
- keep static select allow-list validation unchanged
- cover single and multiple dynamic selects with numeric provider keys

## Reproduction

EPG Enricher v1.13.5 defines a required dynamic select action field named `playlist_id`. Its options provider returns numeric playlist model IDs. The host normalizes those option keys for Filament, but Livewire submits a selected numeric key such as `5` as integer `5`. The action payload then fails the host `string` validation with `The playlist id field must be a string.` before the plugin runs, even though Health Check succeeds and the selected playlist is valid.

## Testing

- `REVERB_APP_SECRET=test php artisan test --compact --filter="accepts integer values from numeric dynamic select option keys" tests/Feature/PluginSystemTest.php`
- `REVERB_APP_SECRET=test php artisan test --compact tests/Feature/MakePluginCommandTest.php tests/Feature/PluginDeclaredTableUiTest.php tests/Feature/PluginMalwareScannerTest.php tests/Feature/PluginSettingsSchemaGroupingTest.php tests/Feature/PluginSystemTest.php tests/Feature/PluginUpdateCheckerTest.php tests/Feature/PluginsDashboardTest.php`
- `git diff --check`

### DIFF
--- a/app/Plugins/PluginSchemaMapper.php
+++ b/app/Plugins/PluginSchemaMapper.php
@@ -3,6 +3,7 @@
 namespace App\Plugins;
 
 use App\Models\Plugin;
+use Closure;
 use Filament\Forms\Components\Select;
 use Filament\Forms\Components\TagsInput;
 use Filament\Forms\Components\Textarea;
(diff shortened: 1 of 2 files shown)


sparkison:
Looks good!

PR #1280, opened 2026-07-12
feat: add Telegram alerts channel

### LINKED ISSUE #1218: Alerts: Telegram notification service
## Discussion

Add Telegram as a configurable notification service under Alerts so m3u-editor can notify users when relevant events happen.

This should be discussed as part of the Alerts architecture rather than as a standalone one-off integration.

## Questions to Decide

- Should Telegram be configured globally, per user, per alert rule, or a combination of those?
- Should it support direct chats, groups, or both?
- How should chat IDs be discovered or validated?
- Should messages be sent immediately only, or also support grouped/digest notifications?
- Which alert events should be eligible for Telegram delivery first?

## Suggested Direction

- Add a Telegram notification channel/provider under Alerts.
- Store the required bot token and chat target configuration securely.
- Allow alert rules to select Telegram as a delivery target.
- Provide a test notification action so users can verify setup before enabling rules.
- Keep message templates concise and include enough context to identify the affected playlist, stream, DVR item, or job.

## Acceptance Criteria for a Future Implementation

- Users can configure Telegram as an Alerts notification service.
- Users can send a test Telegram notification from the configuration UI.
- Alert rules can target Telegram recipients.
- Notification failures are visible in logs or alert delivery history.
- Secrets are not exposed back to the UI after save.

tjelite1986:
## Summary

Implements #1218: adds **Telegram** as a third alert channel alongside the existing Discord and Slack integrations, following the approach suggested in the issue — global settings under **Settings > Alerts**, same logic as the other alerts, using [laravel-notification-channels/telegram](https://github.com/laravel-notification-channels/telegram) (v7, supports Laravel 13 / PHP 8.3+).

## How it works

- **Settings**: new `telegram_alerts_enabled`, `telegram_bot_token` and `telegram_chat_id` in `GeneralSettings` + settings migration. Both token and chat ID are stored in settings (no `services.php` config entry needed) — the token is passed per-message via `TelegramMessage::token()` and the chat ID via on-demand notification routing.
- **`AlertService`**: sends to Telegram when the toggle is on and both token and chat ID are set; `isEnabled()` now counts Telegram as an active channel (so the `Additional Notifications` opt-ins and the log handler work with Telegram-only setups).
- **`TelegramAlert` notification**: queued (`ShouldQueue`), mirroring the queued delivery of the Spatie Discord/Slack alert jobs. Messages are sent **without a parse mode** (`->normal()`) so forwarded log content (underscores, brackets, JSON context) can never break Telegram's Markdown entity parsing and silently drop an alert.
- **Preferences UI**: Telegram section in the Alerts tab with an enable toggle, a step-by-step setup guide (BotFather + `getUpdates`), a masked/revealable token input, chat ID input, and a "Send test alert" header action (sent synchronously via `notifyNow` for immediate feedback), mirroring the Discord/Slack sections.
- **Loop guard**: `AlertOnJobFailed` now skips permanently-failed `TelegramAlert` deliveries, so a misconfigured Telegram channel can't generate alerts about its own failed alerts in an endless loop.

## Screenshot-free UI summary

The Telegram section sits between Slack and Additional Notifications in the Alerts tab and follows the exact same layout conventions (toggle → setup guide → inputs, test action in the section header, all gated on the enable toggle).

## Tests

New `tests/Feature/TelegramAlertTest.php` (8 tests, 13 assertions) covering:

- alert routed to the configured chat ID when enabled
- no send when disabled, or when token/chat ID is missing (dataset)
- `isEnabled()` only true when fully configured
- `toTelegram()` payload: plain text (no `parse_mode`), correct token
- `AlertOnJobFailed` forwards other job failures but skips failed `TelegramAlert` deliveries (loop guard)

Local verification: new test file passes; full suite `1607 passed, 29 skipped (4264 assertions)` — the only 3 failures are the known env-only `PlayerPopoutRouteTest` Vite-manifest failures (assets not built in my clone; CI builds them). Pint clean on all touched files. `composer.lock` diff contains only the new package.

Closes #1218

### DIFF
--- a/app/Filament/Pages/Preferences.php
+++ b/app/Filament/Pages/Preferences.php
@@ -19,6 +19,7 @@ use App\Models\PlaylistAlias;
 use App\Models\StreamFileSetting;
 use App\Models\StreamProfile;
 use App\Notifications\Notification as AppNotification;
+use App\Notifications\TelegramAlert;
 use App\Rules\Cron;
 use App\Rules\ValidDateFormat;
 use App\Services\DateFormatService;
@@ -57,6 +58,7 @@ use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Facades\Mail;
+use Illuminate\Support\Facades\Notification as NotificationFacade;
 use Illuminate\Support\Facades\Redis;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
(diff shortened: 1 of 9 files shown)


sparkison (commented):
Looks good!

PR #1289, opened 2026-07-15
feat: DNS failover support for playlist aliases

Warbs816:
## Summary

Playlists already handle DNS failover nicely (fallback URLs + automatic promotion when one goes down), but aliases don't — after a provider DNS change, alias upstream URLs stay broken until they're edited by hand. This PR brings failover to playlist aliases with a per-alias **DNS failover mode**:

- **Static** (default) — exactly the current behavior; existing aliases are untouched.
- **Inherit from source playlist** — the alias follows the source playlist's current DNS, including its fallback list. When the playlist fails over, the alias picks up the new URL automatically while keeping its own username/password. Works for both standard and custom-playlist aliases.
- **Independent failover URLs** — each credential entry gets its own list of fallback URLs with promote-on-failure, mirroring the playlist behavior.

## How it works

- **Inherit** resolves at read time: the alias `xtream_config` accessor injects the source playlist's live URL into each entry, so a playlist failover propagates to the alias instantly with no extra jobs or state.
- **Independent** health checks run through existing paths only (no new scheduler): the alias stats refresh self-heals on failure, and when a parent playlist promotes a fallback URL it queues a health check for its independent-mode aliases. Each fallback URL also gets a test button in the form.
- Credential entries now carry the source `playlist_id` they belong to. This also fixes a latent issue on multi-provider aliases: entries were matched to sources by URL, so after a source playlist failed over to a new URL, the wrong provider's credentials could be selected. Matching is now by playlist id first, with URL matching kept for existing configs.

## Changes

- New `dns_failover_mode` column on `playlist_aliases` (default `static`) + `DnsFailoverMode` enum
- `PlaylistAlias`: inherit-mode URL resolution in the `xtream_config` accessor, `promoteXtreamUrl()` for independent mode, entry lookup by playlist id, and URL matching that includes fallback URLs (keeps M3U-source URL rewriting working after a promotion)
- `XtreamHealthService`: reusable per-entry URL probing (`findWorkingEntryUrl` / `resolveWorkingAliasUrls`)
- New `CheckPlaylistAliasDnsHealth` job, dispatched from `Playlist::promoteXtreamUrl()` for independent-mode aliases
- `UpdateXtreamStats`: on a failed alias stats fetch in independent mode, probes fallbacks, promotes a working URL, and retries once
- Alias form: failover mode select, per-entry fallback URL repeater with test actions, inherit-aware URL field
- Trailing-slash normalization for alias URLs now covers fallback URLs and works on the stored config shape (the previous hook read a key that never exists on the list-shaped config)
- 21 new tests covering all three modes, promotion semantics, healing paths, and legacy config handling

## Test plan

- [x] Existing aliases behave exactly as before (mode defaults to Static)
- [x] Inherit mode: fail the source playlist over to a fallback URL and confirm the alias streams/API follow it, with the alias's own credentials
- [x] Independent mode: add fallback URLs to an entry, take the primary down, and confirm the alias promotes a working URL (stats view or parent playlist failover both trigger it)
- [x] Custom-playlist alias with entries from multiple source playlists picks the right credentials after a source playlist failover
- [x] Editing and re-saving an alias preserves its fallback URLs and entry–playlist links


sparkison:
I'm going to close this in favor of a much simpler solution 😉 I'm going to take advantage of the existing `promoteXtreamUrl` function that is called when a promotion happens and tie it in there. This is how we're handling EPG's already, so makes sense to cover aliases here as well. If we need something more feature rich after feel free to open another PR, but I think what I'm implementing now will cover it...

Warbs816:
Happy with that, thanks very much!

PR #1291, opened 2026-07-15
fix(media-server): single-authority seek + byte-range fast-seek for Emby broadcasts

Grimothy:
## What

Closes three subtitle/seek defects in the Emby/Jellyfin network broadcast path that all surface when a user seeks mid-programme:

1. **Subtitle desync after seek** — dual seek engines (media server seeks video via `StartTimeTicks`; proxy seeks subtitle via `-ss N -itsoffset -N`) drifted up to a full seek-second apart. Switched to a single media-server seek that uses Emby's `/Subtitles/{i}/{ticks}/Stream.{fmt}` URL to rebase subtitle cues to zero at the seek point, and pass `subtitle_seek_seconds=0` to the proxy so it never re-seeks the subtitle input.

2. **VideoCodec=copy remux starts at byte 0 when seeked** — Emby serves a VideoCodec=copy remux URL with `Accept-Ranges: none` and unknown duration, so client-side `-ss` can't seek it. Detect the remux in `getDirectStreamUrl` and rewrite it to the `static=true` variant which is seekable.

3. **HEVC byte-seek hangs** — Plex/Emby direct file endpoints with byte-seek on HEVC content hang the transcode. Added a `#range=N-` byte offset calculated from runtime ticks → byte size, falling back to no byte-seek when size can't be resolved.

## How

- `EmbyJellyfinService::getSubtitleUrl($itemId, $seekSeconds=0)` returns `{url, language, server_seeked}` — `server_seeked=true` means cues already rebased to 0, so the proxy should leave the subtitle input untouched. `getStreamByteSize()` returns `[bytes, duration_seconds]` for fast-seek math.
- `NetworkBroadcastService::startViaProxy()` builds `subtitle_url`/`subtitle_language`/`subtitle_seek_seconds`/`audio_stream_index` payloads, decides between server-side media-server seek vs proxy-side `-ss` based on whether the URL is a remux, and computes `#range=` from byte size when applicable.
- `NetworkHlsController::variant()` (route + method surgically extracted from a larger networks admin commit) serves the sub-playlist + `.ts`/`.vtt` segments referenced from a master playlist when subtitles are enabled.
- `MediaServerService::getAudioStreamIndexForLanguage()` interface method resolves the audio stream index for a given language preference so `computeNextStreamConfig()` can pre-resolve for the next programme before the auto-transition fires (otherwise the next programme would have no audio preference because it hasn't been built yet).

## Schema

No migrations in this PR. Uses the columns upstream already shipped:
- `preferred_audio_track` — audio stream selection by language code or stream ID/index
- `preferred_subtitle_track` — proxy subtitles_enabled toggle is derived from `!empty($network->preferred_subtitle_track)` (any preference means operator wants subtitles)

## Tests

55/55 passing across 7 feature test files: `NetworkBroadcastPayloadTest`, `NetworkBroadcastTransitionTest`, `EmbySubtitleUrlTest`, `NetworkHlsSubtitleVariantTest`, `PlexTrackPreferenceTest`, `EmbyJellyfinTrackPreferenceTest`, `NetworkBroadcastSeekTest`. The upstream track-prefs tests continue to pass — this PR is orthogonal to upstream's track-prefs work and only adds the seek/desync/HLS-variant plumbing around it.

## Out of scope

- Audio track selection is already covered upstream (`PlexService::resolvePreferredStreamId` / `EmbyJellyfinService::resolvePreferredStreamIndex`). This PR drops the original PHP-side audio stream index resolution that duplicated upstream's work and keeps only the unique pre-resolve for auto-transition.
- Admin UI for audio/subtitle preferences is upstream's `NetworkResource` fields.
- Network admin UI / schedule regeneration / HLS controller admin side is tracked in a separate PR (`chore/network-admin-checkpoint-cleanup`).

### DIFF
--- a/app/Http/Controllers/NetworkHlsController.php
+++ b/app/Http/Controllers/NetworkHlsController.php
@@ -70,15 +70,22 @@ class NetworkHlsController extends Controller
 
             $playlist = $response->body();
 
-            // Rewrite segment URLs to go through our proxy route
-            // FFmpeg outputs segment names like "live000001.ts" in the playlist
-            // We need to rewrite them to full URLs: /m3u-proxy/broadcast/{uuid}/segment/live000001.ts
-            $baseUrl = url("/network/{$network->uuid}");
-            $playlist = preg_replace(
-                '/^(live\d+\.ts)$/m',
-                $baseUrl.'/$1',
-                $playlist
-            );
+            if ($this->isMasterPlaylist($playlist)) {
+                // Subtitles are enabled for this network: the proxy returns a master
+                // playlist referencing a video variant + subtitle variant playlist
+                // instead of a flat list of .ts segments.
+                $playlist = $this->rewriteMasterPlaylist($playlist, $network);
+            } else {
+                // Rewrite segment URLs to go through our proxy route
+                // FFmpeg outputs segment names like "live000001.ts" in the playlist
+                // We need to rewrite them to full URLs: /m3u-proxy/broadcast/{uuid}/segment/live000001.ts
+                $baseUrl = url("/network/{$network->uuid}");
+                $playlist = preg_replace(
+                    '/^(live\d+\.ts)$/m',
+                    $baseUrl.'/$1',
+                    $playlist
+                );
+            }
 
             return response($playlist, 200, [
                 'Content-Type' => 'application/vnd.apple.mpegurl',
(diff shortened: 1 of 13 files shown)


sparkison:
Think I have the direction now! Between what we have, this PR, and the upcoming updates to the proxy, hopefully it wraps this all up nicely 🤞

PR #1292, opened 2026-07-15
fix(dvr): daily deep scan so airings beyond 30-min window get scheduled

Grimothy:
## Problem

Recording rules were not scheduling new airings from EPG data added after the rule was created. The rule's "matched airings" UI panel showed programmes 14 days ahead, but no `dvr_recordings` rows were being created for any of them — only airings inside a 30-minute window were scheduled.

### Root cause

| Path | Lookahead window |
|---|---|
| Rule's "matched airings" panel | 14 days |
| `scheduleRuleImmediately()` (on create/enable) | 14 days |
| Per-minute `DvrSchedulerTick` | **30 minutes** |

So once a rule was created, its 14-day scan only captured EPG data present at that moment. Anything the EPG refresh added in the following hours (XMLTV, Schedules Direct, etc.) sat unscheduled until it slid into the 30-minute tick window. For airings hours/days in the future that never happened — they just stayed invisible to the schedule.

## Fix

Move the EPG match step out of the per-minute tick and into a new **daily deep scan** with the full `initial_lookahead_days` window.

- **`DvrSchedulerTick`** (every minute) now only handles trigger/stop of already-scheduled recordings. Minute-precision for actual recording start/stop is preserved.
- **`DvrDeepScan`** (new, daily at `dvr.deep_scan_hour`, default 03:00) calls `matchAndSchedule` with the 14-day `initial_lookahead_days` window. Picks up new EPG data added between scans.
- **Rule create/re-enable** still triggers `scheduleRuleImmediately()` for instant UX feedback — unchanged.

## Latency profile

| Path | Latency |
|---|---|
| Create rule | Seconds |
| New EPG data added | ≤24h (daily scan) |
| Recording start | Minute precision |

## Future work (TODO in `DvrDeepScan`)

Replace the daily cadence with an EPG-event-driven scan that fires when XMLTV/Schedules Direct imports add new programmes. The daily scan then becomes a belt-and-suspenders fallback.

## Files changed

- `app/Services/DvrSchedulerService.php` — stripped match step + cache throttle from `tick()`; exposed `matchAndSchedule` as public
- `app/Jobs/DvrDeepScan.php` — new daily job
- `routes/console.php` — registered daily schedule
- `config/dvr.php` — replaced `scheduler_match_interval_minutes` / `scheduler_lookahead_minutes` with `deep_scan_hour`
- `tests/Feature/DvrSchedulerServiceTest.php` — refactored `tick()` → `matchAndSchedule(30)` for scheduling tests, restored `tick()` for trigger/stop tests, added 4 new architecture tests

### DIFF
--- a/app/Jobs/DvrDeepScan.php
+++ b/app/Jobs/DvrDeepScan.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace App\Jobs;
+
+use App\Services\DvrSchedulerService;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Queue\Queueable;
+use Illuminate\Support\Facades\Log;
+
+/**
+ * DvrDeepScan — Match all enabled DVR rules against the EPG lookahead window.
+ *
+ * Runs on a daily schedule (default 03:00, configurable via dvr.deep_scan_hour).
+ * Picks up new EPG data added since the previous run so airings more than 30
+ * minutes from now are still scheduled. The per-minute tick only handles
+ * trigger/stop of already-scheduled recordings; this is the steady-state
+ * rescan that catches EPG refreshes between user actions.
+ *
+ * TODO: replace this daily cadence with an EPG-event-driven scan that fires
+ * when XMLTV or Schedules Direct imports add new programmes. The daily scan
+ * will then become a belt-and-suspenders fallback.
+ */
+class DvrDeepScan implements ShouldBeUnique, ShouldQueue
+{
+    use Queueable;
+
+    public int $tries = 1;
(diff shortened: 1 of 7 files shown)


sparkison:
great direction! Moving the scheduler out of the tick 😉 was wondering about this and didn't get a chance to get back in and review, glad to know this will work though, should help quite a bit with load. I just found the once/manual rules didn't work so added a fix for that quick.

Grimothy:
> great direction! Moving the scheduler out of the tick 😉 was wondering about this and didn't get a chance to get back in and review, glad to know this will work though, should help quite a bit with load. I just found the once/manual rules didn't work so added a fix for that quick.

Thanks for the kind words -- About the once rules maybe thats why the tv apps arent working properly with recordings

PR #1296, opened 2026-07-16
feat: add authenticated Xtream request API

### LINKED ISSUE #1197: APP integrated request features

A lot of platforms lack the ability to **integrate into sonarr and radarr** for content downloads and integrations. Currently **M3U-editor at the web server level does not have a feature to support this but I plan on building one**. I am **requesting** that the primary developers of the mobile, desktop and tvos applications consider **adding** **increased functionality and preferential** treatment of playlists hosted from M3U-editor.

A separate **navigational element** that would **allow for the searching and requesting of content that is not available on the platform**. I understand that this is in all likely hood a very fringe ask. For individual that bring their own content this will allow a centralized platform for controlling content.

In terms of the order of operations, I would suspect that I would need to build out the server elements first allowing you to connect to the appropriate api endpoints to facilitate the integration.

Just using multiple available open source solutions like Jellyseer

**Additional context**

1. Main server features need to be built out and tested
2. Application adoption from the developers

Serph91P:
## Summary

- add a versioned, authenticated Xtream Request API for search, submit, history, status, and dismiss actions
- scope request integrations and history to the authenticated PlaylistAuth and effective playlist
- reuse existing Sonarr/Radarr services and media request models with normalized, secret-free responses
- advertise only available content types and normalize series seasons for client selection
- cover authorization, validation, rate limiting, outages, deduplication, idempotency, approval modes, status, and dismiss behavior

## Verification

- `php artisan test --compact tests/Feature/XtreamRequestActionsTest.php tests/Feature/PlaylistAuthMaxConnectionsTest.php`: 62 passed, 295 assertions on the refreshed exact head
- migration behavior is covered in `XtreamRequestActionsTest.php` on SQLite and PostgreSQL CI
- exact-head GitHub CI Tests: 2,573 passed, 8,694 assertions, 28 skipped
- exact-head Code Quality, Security Scan, Tests, and Docker Build Validation: successful
- `npm run build`
- `vendor/bin/pint --test`
- Composer validation and audits, PHP syntax, checkpoint, plugin validation, diff, conflict-marker, Unicode, and secret-hygiene gates are covered by the exact-head CI workflow
- an extra local full Pest run without the required Redis service reached 2,142 passed and 28 skipped, with 431 environment failures caused by unavailable Redis at `127.0.0.1:36790`; the exact-head CI run with Redis completed successfully

Fixes #1197


Serph91P:
Review blockers on exact head `7b2d95ca0c9240379ad574ff29e9c818ba62763c`:

1. P1: `database/migrations/2026_07_16_174950_add_active_request_unique_index_to_media_requests_table.php:14-18` does not create the intended partial unique index. Laravel's schema grammars ignore `where()` on this index definition. The actual SQLite DDL is `CREATE UNIQUE INDEX "media_requests_active_unique" on "media_requests" ("playlist_auth_id", "arr_integration_id", "external_id", "request_type")`, with no status predicate, and Laravel's PostgreSQL `compileUnique()` path also emits no predicate. This makes completed and rejected history unique too, can reject valid later requests, and can make the migration fail when historical duplicates already exist. Please create a real driver-appropriate active-only uniqueness guard and add coverage proving completed or rejected rows do not block a new request while pending or approved rows do.

2. P2: `app/Services/ContentRequestService.php:302-316` still falls back to title matching whenever a queue item has no `externalId`. That can report a different same-title item's queue state and contradicts the remediation requirement that status correlation use stable content identity only. Identity-less queue entries should not match an externally identified request. Please add the missing-identity counterexample alongside the different-external-ID case.

The review base also drifted during this pass. Current `upstream/experimental` is `693166216abe2ae19e1c4627ab4f193fbaa9e5b9`, which is not an ancestor of this PR head. The successful GitHub checks are tied to base `b1ede2a01f3788e4d64b014d704001e81f09fcd8`, not the current integration head. Please incorporate the current base and rerun the exact-head checks after fixing the blockers.

Serph91P (commented):
Reviewed the migration hardening at f76dbcac. Duplicate reconciliation now preserves approved rows first and otherwise keeps the newest requested_at, with id used only as a deterministic tie-breaker. I also verified the reconciliation and partial unique index on SQLite and PostgreSQL 16, including historical and identity-less rows. The focused tests, full suite, Pint, frontend build, and current-head CI are green. No blocking findings from me. I cannot submit a formal approval because this account owns the pull request.

Grimothy:
Heads up since you're in XtreamApiController — there's a related bug on schedule_dvr (around line 2890). It creates a DvrRecordingRule with type=Once + manual_start/manual_end + no programme_id. The scheduler routes by type, so matchOnceRule falls through to matchOnceRuleViaDummyEpg which doesn't match without dummy_epg on the playlist — the action returns success but no DvrRecording ever fires. Fix is one line: switch the type to Manual, since that's what the rule's data actually represents (matchManualRule already reads manual_start/manual_end properly).

Putting it up as a draft PR against experimental in the next day or two. Wanted to flag in case you want to coordinate or fold it into your PR. I'm happy to either keep them separate or merge — your call.

Grimothy:
Follow-up: opened as a draft — https://github.com/m3ue/m3u-editor/pull/1299. Changes `type=Once` → `Manual` in schedule_dvr and extends the boot event so Manual rules fire `scheduleRuleImmediately` (currently it only fires for Series, which would have made the new Manual rule wait for the daily deep scan). Happy to fold into your PR if you'd rather have it all in one — just say the word.

Grimothy:
Quick update — the maintainer (sparkison) folded the Once/Manual fix into PR #1292 (merged today as commit `8a374a3a` — "Fix once/Manual rule creation flows broken"). That broader fix changes `DvrRecordingRule::boot()` to fire `scheduleRuleImmediately` for *all* rule types instead of just Series.

I've trimmed PR #1299 accordingly — it now only carries the one-line `XtreamApiController` change (`type=Once` → `type=Manual` at line 2890) plus the test coverage. The boot-event change is no longer needed; PR #1292 covers it. The Xtream-specific fix is still relevant because the type tag the endpoint emits must match the data shape (manual_start/manual_end + no programme_id → Manual), independently of how scheduling is dispatched downstream.

Serph91P (commented):
Accepted after independent review at ca5b50ec.

No blocking findings. I verified the upstream refresh has no conflict artifacts, the focused request coverage passes with 62 tests and 295 assertions, and the PostgreSQL 16 migration fixture preserves approved requests, otherwise keeps the newest requested_at row, rejects surplus active rows, and enforces the active-only partial unique index. Pint, dependency audits, frontend build, PHP syntax, plugin checks, and diff integrity checks are clean. All four GitHub checks for this exact head are successful.

The PR remains Draft as required.

Serph91P:
@sparkison This is ready for review and merge.

PR #1309, opened 2026-07-19
Fix max connections defaulting to 1 when proxy is disabled

Leonard013:
This PR resolves an issue where fully disabling proxy features (e.g., when the proxy integration is globally disabled via configuration or permissions) causes the max connections / tuner count for all newly created playlists to default to `1`, regardless of whether the provider supports more connections.

### Root Cause
1. In `PlaylistResource.php` (and other resources), the `streams` TextInput default value of `0` is ignored and omitted when the form section containing it is hidden (e.g., because `canUseProxy()` evaluates to `false`).
2. When the value is omitted during playlist creation, the database column default of `1` (defined in the `2025_03_20_131924_add_hdhr_tuners_to_playlists` migration) is saved.
3. The Xtream API and M3U/HDHomeRun generation endpoints contain fallback logic to retrieve `max_connections` from `xtream_status` if `$playlist->streams` is `0`. However, since it is saved as `1` in the database, this fallback is never triggered, forcing the tuner count / max connections to `1`.
4. In `PlaylistGenerateController.php` (HDHomeRun discover endpoint), an undefined variable `$xtreamStatus` was referenced, causing the fallback to fail when resolving tuner count.

### Solution
1. Add a database migration changing the default value of the `streams` column to `0` for the `playlists`, `custom_playlists`, and `merged_playlists` tables, and migrate existing records with the default value of `1` to `0` to enable the provider max connections fallback.
2. Set the default attribute value `'streams' => 0` on the `Playlist`, `CustomPlaylist`, and `MergedPlaylist` models to ensure new instances default to `0` even if column defaults aren't processed immediately by the application.
3. Define the missing `$xtreamStatus = $playlist->xtream_status;` variable in `PlaylistGenerateController::getDeviceInfo()` so the fallback retrieves the correct provider `max_connections` for HDHomeRun.
4. Add feature tests covering new playlist default values, Xtream API, and HDHomeRun tuner counts under disabled proxy configurations.

### DIFF
--- a/app/Http/Controllers/PlaylistGenerateController.php
+++ b/app/Http/Controllers/PlaylistGenerateController.php
@@ -683,6 +683,7 @@ class PlaylistGenerateController extends Controller
 
         // Return the HDHR device info
         $uuid = $playlist->uuid;
+        $xtreamStatus = $playlist->xtream_status;
         $tunerCount = (int) $playlist->streams === 0
             ? ($xtreamStatus['user_info']['max_connections'] ?? $playlist->streams ?? 1)
             : $playlist->streams;
--- a/app/Models/CustomPlaylist.php
+++ b/app/Models/CustomPlaylist.php
@@ -48,6 +48,15 @@ class CustomPlaylist extends Model
         'processing_config' => 'array',
     ];
 
+    /**
+     * The model's default values for attributes.
+     *
+     * @var array
+     */
+    protected $attributes = [
+        'streams' => 0,
+    ];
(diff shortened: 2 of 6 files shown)


sparkison:
Appreciate the contrib! 🙏 we've already got fixes in place for this though 😉 the standard Playlist keeps these fields exposed without proxy (out in release `0.11.67`), and the remaining playlist types will get this with the next release (I missed it in my first pass)

PR #1313, opened 2026-07-21
fix (ui): Scrollbar disappear after executing a bulk action operation

thanhvle94:
## Issue
The scrollbar on the main page in Live Channels > Channels disappears after executing any bulk action. The only way to restore it is to refresh the page to re-enable it.

## Reproduction
Live Channels -> Channels (More than 25 channels) -> Select any -> Bulk Actions -> Execute any actions 

## Test
Execute the same procedure and observe that the scrollbar can scroll again.

## Note
PHP directly from Filament
https://github.com/filamentphp/filament/blob/4.x/packages/support/resources/views/components/modal/index.blade.php

### DIFF
--- a/resources/views/vendor/filament/components/modal/index.blade.php
+++ b/resources/views/vendor/filament/components/modal/index.blade.php
@@ -1,0 +1,283 @@
+@php
+    use Filament\Support\Enums\Alignment;
+    use Filament\Support\Enums\SlideOverPosition;
+    use Filament\Support\Enums\Width;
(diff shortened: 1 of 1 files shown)


sparkison:
Nice find! This was actually resolved in the vendor lib upstream in v5.7.1, so I'm going to update instead to get the official fix 😉 

Here's the upstream fix that will be included in the next release: https://github.com/filamentphp/filament/commit/8f9bc361b0b844dcddae5831210c4bdeab56fea9

thanhvle94:
Awesome!

PR #1316, opened 2026-07-21
feat: add filter to live, VOD, and series processing

thanhvle94:
Filter by selected groups or categories so you can quickly uncheck them. This is especially handy when managing massive lists.

image


sparkison:
Looks good! Nice addition 😉

PR #1343, opened 2026-07-28
fix(alias): resolve alias ownership through the effective playlist

ethan-manny:
## Problem

`PlaylistAliasPolicy` resolves the alias owner through `$playlistAlias->playlist->user_id` in all six methods. `playlist` is the `belongsTo` for a **standard** playlist, so it is `null` on an alias of a **custom** playlist. Reading `->user_id` on null raises a PHP warning, which Laravel promotes to an `ErrorException`:

```
Attempt to read property "user_id" on null
at app/Policies/PlaylistAliasPolicy.php:40
```

`User::canAccessPanel()` returns `true` for everyone, so non-admin users do reach the resource. `$user->isAdmin()` short-circuits the expression, so **admins never see this** — which is why it has gone unnoticed. A non-admin cannot view, edit or delete their own alias when it points at a custom playlist.

Reproduced in the browser on `dev` with a non-admin user who owns a custom-playlist alias.

## Fix

Resolve through the existing `getEffectivePlaylist()`, which returns the standard playlist or the custom one. Behaviour for standard-playlist aliases is byte-for-byte unchanged; the six duplicated expressions collapse into one private helper.

```php
private function owns(User $user, PlaylistAlias $playlistAlias): bool
{
    return $user->isAdmin() || $user->id === $playlistAlias->getEffectivePlaylist()?->user_id;
}
```

The null-safe operator matters independently: an alias whose target playlist has been deleted now denies access rather than erroring.

## Tests

`tests/Feature/PlaylistAliasPolicyTest.php` covers all five abilities against both alias types:

- a non-admin owner is granted every ability on a custom-playlist alias
- a non-admin owner is granted every ability on a standard-playlist alias (no regression)
- a non-owner is still denied
- an admin is still granted regardless of target
- a non-admin owner can open the alias edit page

**11 of the 21 fail on the current policy** and pass with the fix.

## Notes

- No user-facing strings, so no `lang/` changes.
- `vendor/bin/pint --test` passes.
- `php artisan checkpoint:scan` reports no findings introduced by this branch (verified by diffing finding IDs against `dev`; the scan is already non-zero on `dev` for pre-existing issues).

### DIFF
--- a/app/Policies/PlaylistAliasPolicy.php
+++ b/app/Policies/PlaylistAliasPolicy.php
@@ -20,8 +20,7 @@ class PlaylistAliasPolicy
      */
     public function view(User $user, PlaylistAlias $playlistAlias): bool
     {
-        // PlaylistAlias doesn't have direct user_id, check through playlist relationship
-        return $user->isAdmin() || $user->id === $playlistAlias->playlist->user_id;
+        return $this->owns($user, $playlistAlias);
     }
 
     /**
@@ -37,7 +36,7 @@ class PlaylistAliasPolicy
      */
     public function update(User $user, PlaylistAlias $playlistAlias): bool
     {
-        return $user->isAdmin() || $user->id === $playlistAlias->playlist->user_id;
+        return $this->owns($user, $playlistAlias);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sparkison:
Nice catch! Reviewing now...

PR #1344, opened 2026-07-28
feat(alias): support channel filters on custom playlist aliases (#1325)

### LINKED ISSUE #1325: Playlist alias channel filter for custom playlists
When creating a playlist alias of a custom playlist there is no option for channel filters.

A clear and concise description of what you want to happen:
Creating an alias of a standard or custom playlist should allow the user to use the channel filter functionality as is currently implemented for standard playlists.

ethan-manny:
Closes #1325

## Problem

Creating a playlist alias of a **custom** playlist offered no channel filter. The filter existed only for standard playlists, in two places:

- `PlaylistAliasResource` hid the *Channel Filter* fieldset unless `playlist_id` was set.
- `PlaylistAlias::channels()` / `series()` returned early for the custom branch, before any filter was applied.

## Approach

Custom playlists group content with per-playlist Spatie tags rather than provider groups. Content carrying **no** tag falls back to its provider group/category name in the generated output — and the Xtream category endpoints already list both. So the filter matches on *effective display name*: the group tag if the content has one, otherwise the provider name.

That means the picker has to offer both. A tag-only list would be empty for a custom playlist whose channels were never assigned to a custom group, making the feature unusable there; a provider-only list would omit user-created groups, which have no `Group` row at all.

To list both in a single searchable, paginated table, `CustomPlaylist` builds a `UNION` of the two name sources, wrapped by `CustomPlaylistGroup` — a small table-less model keyed by **name**. Keying on name means the picker's values are already the format `group_filter` stores, so unlike the standard-playlist selects no id↔name translation is needed.

JSON tag names are read with Laravel's portable `name->en` syntax, which compiles to `json_extract(...)` on SQLite and `->>` on PostgreSQL.

## Changes

| File | |
|---|---|
| `app/Models/CustomPlaylistGroup.php` | **New.** Derived, table-less model keyed by name |
| `app/Models/CustomPlaylist.php` | `filterableGroupsQuery()` / `filterableCategoriesQuery()` build the union |
| `app/Models/PlaylistAlias.php` | Applies the filter in the custom branch of `channels()` / `series()` |
| `app/Filament/Tables/CustomPlaylistGroupsTable.php`, `CustomPlaylistCategoriesTable.php` | **New.** Picker tables mirroring `SourceGroupsTable` |
| `app/Filament/Resources/PlaylistAliases/PlaylistAliasResource.php` | Modal-table pickers matching the standard-playlist experience, plus a note that the lists combine both sources |
| `app/Http/Controllers/XtreamApiController.php` | Filters the custom-playlist live/VOD/series category listings |
| `app/Filament/GuestPanel/Resources/{Vods,Series}` | Guest panel honours the filter |

Enforcing the filter on the `channels()` / `series()` relationships means M3U generation, EPG output, the Xtream API and channel counts all inherit it without per-call-site wiring.

Storage format is unchanged (`selected_groups` / `selected_vod_groups` / `selected_categories` as name lists), so existing aliases and `hasGroupFilter()` are untouched.

## Custom live group ordering

*Sort groups in custom order* now works for custom-playlist aliases too, not just standard ones. The standard implementation ranks on `channels.group_internal`, which is meaningless for a custom playlist, so ordering here ranks on the group the client is actually shown — the channel's custom tag, or the provider group name it falls back to when untagged. The JSON tag name is extracted by the query grammar so it stays portable.

The form controls simply drop their `playlist_id` gate: the existing sort helpers already degrade correctly without a playlist id, because custom-playlist selections are stored as names and need no id/name resolution.

## Included behaviour change

`getChannelQuery()` already resolved `custom_group_name` for an alias of a custom playlist, but the M3U writer only applied it when `$type === 'custom'`. An alias emitted the **provider** group name while the Xtream API, for that same alias, reported the **custom tag** — the two outputs disagreed, and the user's own grouping was ignored in M3U. Without this, ordering by tag while labelling by provider group produces incoherent output.

Fixed by gating on the `$isCustomContext` flag computed two lines above. **This changes M3U output for existing custom-playlist aliases**: groups now carry the custom tag name rather than the provider group name. It is a separate commit if you'd prefer it split out.

## Tests

`tests/Feature/PlaylistAliasCustomPlaylistFilterTest.php` — 36 tests covering tag matching, the untagged provider-name fallback, tag-beats-provider-group, live/VOD independence, series categories, the option lists, the Xtream category endpoints, the picker table contents, group ordering (query and M3U output), and the form round-trips.

Three worth calling out, because the obvious version of each was too weak:

- The picker table renders as its own Livewire component, so mounting the modal from the form never runs its query. Those tests drive `TableSelectLivewireComponent` directly, verified to fail when the query is broken.
- A custom playlist with **zero** tags is its own case: the empty tag collection stayed an Eloquent collection and blew up when merged with plain-string fallback names. A test with one tagged channel does not catch it.
- The ordering tests were checked by disabling both the sort and the label fix: 5 of the 6 fail without them.

## Notes

- New user-facing strings added to `lang/en.json` (8 keys).
- `vendor/bin/pint --test` passes.
- `php artisan checkpoint:scan` reports no findings introduced by this branch (verified by diffing finding IDs against `dev`).
- Verified locally on SQLite. The union and JSON extraction were confirmed to compile correctly under the PostgreSQL grammar, but **not executed against a live PostgreSQL** — no Postgres available in my environment, so CI is the first real run on it.


sparkison:
Looks good! 👍

PR #1349, opened 2026-07-30
fix(dvr): broadcast dvr.status to every channel a TV session can be on

Grimothy:
## Summary
- `DvrRecordingStatusEvent` only ever broadcast on the bare `tv.{type}.{uuid}` channel. A TV session logged in via PlaylistAuth credentials (rather than owner_auth) subscribes only to its own `tv.{type}.{uuid}.{authId}` channel — `TvNotificationEvent` already knew this and broadcasts to that channel (plus admin) for every entitled PlaylistAuth. That's why toast notifications worked while the DVR "recording" dot / status label on the TV app never updated from the push alone — it only caught up when scheduling a second recording forced a full `get_dvr_recordings` refresh.
- Extracts the shared channel-set resolution into `App\Events\Concerns\BroadcastsToEntitledTvChannels` so `TvNotificationEvent` and `DvrRecordingStatusEvent` can't drift apart on this again.
- Also fixes the `deleted` status push (`DvrRecording::broadcastDeleted()`), which `AppStateController._onDvrStatusPush` on the TV client relies on to drop cancelled/deleted recordings locally — this is part of m3ue/m3u-tv#131's "reject stale REST results after newer status or delete events" scope, and was previously non-functional for any PlaylistAuth-scoped session.

## Test plan
- [x] `DvrRecordingStatusEventTest`, `DvrTvNotificationTest`, `TvNotificationTest` pass in isolation off `dev` (37 tests)
- [x] Verified live against the test stack with real seeded data: `/api/tv/{user}/{pass}/notifications` for a PlaylistAuth-scoped login reports channel `private-tv.playlist.{uuid}.{authId}`; `DvrRecordingStatusEvent::broadcastOn()` for a recording on that same playlist now includes that exact channel for status-change, cancel, and delete
- [x] Confirmed end-to-end via the live Docker stack: scheduling → recording → cancel → delete all correctly target the PlaylistAuth-scoped channel

### DIFF
--- a/app/Events/Concerns/BroadcastsToEntitledTvChannels.php
+++ b/app/Events/Concerns/BroadcastsToEntitledTvChannels.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Events\Concerns;
+
+use App\Models\PlaylistAuth;
+use Illuminate\Broadcasting\PrivateChannel;
+
+/**
+ * Shared channel-set resolution for TV push events that must reach every
+ * session subscribed under a playlist (owner/admin channel, and one channel
+ * per entitled PlaylistAuth/alias — see PlaylistAuth::scopeEntitledToNotificationRecipient()).
+ *
+ * TvNotificationEvent originated this set; DvrRecordingStatusEvent previously
+ * broadcast only on the bare owner channel, so pushes never reached TV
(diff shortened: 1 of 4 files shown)


sparkison:
Looks good, merging!

PR #1350, opened 2026-07-30
fix(dvr): timestamptz columns store and compare an explicit UTC offset

Grimothy:
## Summary
- `DvrRecordingRule`/`DvrRecording` now set `\$dateFormat = 'Y-m-d H:i:sP'` so Eloquent includes the UTC offset when serializing datetime attributes for storage. Without it, Laravel's default `Y-m-d H:i:s` format drops timezone info, and PostgreSQL's `timestamptz` columns interpreted the bare wall-clock string as UTC — shifting a scheduled time by the full local-vs-UTC offset (e.g. a 12:30pm NY recording landed at 8:30am NY, firing immediately and completing before the user noticed). `DvrSchedulerService` scheduling queries now bind a UTC `nowUtc()` so `scheduled_start <= now` comparisons use the correct absolute instant.
- That format change created a new asymmetry: ad-hoc `where()` comparisons using a bare `Carbon` (`now()`, `\$programme->start_time`, etc.) bind through the query grammar's own offset-less date format, not the model's — an exact-text mismatch against the offset-suffixed stored value. This broke duplicate-recording dedup and failed-recording retry lookups (9 `DvrSchedulerServiceTest` cases: retries created a second row instead of resurrecting the existing Failed one) and made retention's age cutoff fragile on non-Postgres engines. Every such comparison site (`nowUtc()`, new `forCompare()` helper) now formats to match the model's `\$dateFormat` exactly, so read and write agree regardless of DB engine or session timezone.

## Test plan
- [x] `DvrSchedulerServiceTest` (57 tests), `DvrRetentionServiceTest` (9 tests), `XtreamScheduleDvrTest` (7 tests) pass in isolation off `dev`
- [x] Manually verified the timezone fix before→after: `start_time=2026-07-30T16:30:00Z` (12:30pm NY) now stores `manual_start = 16:30:00+00` (correct), not `12:30:00+00` (4h early)

### DIFF
--- a/app/Models/DvrRecording.php
+++ b/app/Models/DvrRecording.php
@@ -22,6 +22,17 @@ class DvrRecording extends Model
     use HasFactory;
 
     /**
+     * Preserve timezone offset when serializing Carbon instances for storage.
+     *
+     * Laravel's default `Y-m-d H:i:s` format discards the timezone, which
+     * breaks `timestamptz` columns — PostgreSQL treats the bare string as UTC
+     * instead of local time, shifting values by the full UTC offset (e.g. +4h
+     * for EDT). Including the offset (`-04:00`) lets PostgreSQL derive the
+     * correct absolute instant.
+     */
+    protected $dateFormat = 'Y-m-d H:i:sP';
+
+    /**
      * @return array<string, string>
      */
     protected function casts(): array
--- a/app/Models/DvrRecordingRule.php
+++ b/app/Models/DvrRecordingRule.php
@@ -19,6 +19,16 @@ class DvrRecordingRule extends Model
     use HasFactory;
 
     /**
+     * Preserve timezone offset when serializing Carbon instances for storage.
+     *
+     * Laravel's default `Y-m-d H:i:s` format discards the timezone, which is
+     * safe for `timestamp without time zone` columns but causes PostgreSQL to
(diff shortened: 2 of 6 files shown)


sparkison:
the timezone/timestamptz issue you found is real, and I confirmed it reproduces exactly as you described (a 12:30pm local recording landing hours off). I ended up going with a different fix in https://github.com/m3ue/m3u-editor/commit/5d824ba5adceb9ef01e435a34feb5669b98224f9, so I'm going to close this PR in favor of that rather than merge here.

The main difference: instead of preserving whatever offset a value happened to carry and reconciling it at each comparison site (nowUtc()/forCompare()), the new approach normalizes every DVR datetime to UTC at the model boundary via a custom cast, so there's only ever one canonical stored representation. That sidesteps a gap in this PR's version — the timestamptz migration only covered Postgres, but this app's default DB_CONNECTION is sqlite, and SQLite has no timezone-aware comparison at all, just raw text ordering. With nowUtc() forcing UTC while the underlying rows could still hold a local offset (inherited from manual_start, which isn't normalized anywhere in the diff), non-UTC-timezone SQLite installs would still hit the same bug this PR set out to fix. Normalizing storage itself avoids needing that reconciliation at every callsite.

Grimothy:
I am not worthy
  :D

PR #1355, opened 2026-07-31
fix(dvr): JSON error responses and one-shot rule cleanup

Grimothy:
## Summary
- `schedule_dvr` could return an HTML error page instead of JSON when the immediate-match path
(recording a program that's already airing) threw — `DvrSchedulerService::scheduleRuleImmediately()`
now catches/logs `matchRule()` failures, matching the existing try/catch already used by the periodic
scheduler tick (`runMatchAndSchedule`).
- `delete_dvr_recording` could do the same for a completed recording with an associated VOD
channel/episode — `DvrRecording`'s `deleting` hook now guards `broadcastDeleted()` and the VOD cascade
block with try/catch, matching the two blocks already guarded right next to them.
- Added a `shouldRenderJsonWhen` rule in `bootstrap/app.php` scoped to the Xtream API routes
(`player_api.php`, `get.php`) so any *other* uncaught exception on these endpoints renders as JSON
instead of Laravel's default HTML error page — these routes are consumed by TV/IPTV clients that always
parse the response as JSON.
- `start-container`: `REDIS_PASSWORD`, `M3U_PROXY_TOKEN`, `REVERB_APP_ID`, and `REVERB_APP_SECRET` are
now loaded from the persisted config volume *before* the generate-if-unset checks run. Previously,
since docker-compose doesn't set these and the persisted `.env` wasn't read until later in the script,
a restarted (not recreated) container would silently regenerate a fresh Reverb secret on every boot.
Any already-connected or cached client would then fail private-channel subscription auth ("Connection
is unauthorized") with no visible error — `dvr.status` (and other) pushes would silently stop arriving
until the app was fully restarted to re-fetch state from scratch.
- **New**: Once/Manual DVR rules are now cleaned up as soon as their recording reaches a terminal
status (Completed/Failed/Cancelled), not just when the recording row is explicitly deleted. Previously,
choosing "Keep recording" to stop an in-progress recording — a common, legitimate choice — left the
spent one-shot rule enabled and orphaned in the editor's Rules list forever, since it would never match
anything again. Series rules are untouched by this change (they still only clean up on deletion, since
a completed episode doesn't mean the season is over).

## Test plan
- [x] `DvrRecordingCascadeTest`, `DvrSchedulerServiceTest`, `XtreamScheduleDvrTest`,
`XtreamCancelDeleteDvrTest` pass
- [x] Reproduced `schedule_dvr` on an already-airing program and `delete_dvr_recording` on a completed
recording with a VOD channel via a local test container — both now return clean JSON instead of an HTML
error page
- [x] Verified the Reverb fix by doing a full container recreate (not just restart) and replaying the
WebSocket private-channel handshake end-to-end — subscription succeeds on the very first boot
- [x] Verified the rule-cleanup fix live: scheduled a recording, cancelled with keep-footage, confirmed
the associated rule was deleted once the recording settled to Completed (previously stayed enabled
indefinitely)

### DIFF
--- a/app/Models/DvrRecording.php
+++ b/app/Models/DvrRecording.php
@@ -71,7 +71,13 @@ class DvrRecording extends Model
             // Broadcast first, before any cascade/file cleanup below can fail —
             // the TV app needs to hear about a deletion regardless of whether
             // the on-disk cleanup succeeds.
-            $recording->broadcastDeleted();
+            try {
+                $recording->broadcastDeleted();
+            } catch (\Throwable $e) {
+                Log::warning("DvrRecording deleting hook: could not broadcast deletion: {$e->getMessage()}", [
+                    'recording_id' => $recording->id,
+                ]);
+            }
 
             // Delete the physical file from disk using the storage facade (file_path is relative).
             if ($recording->file_path) {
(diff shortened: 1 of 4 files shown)


Grimothy:
opps sorry sbout that

Grimothy:
I think without it the DVR in the app doesnt register that a recording has started or any pending status updates. I will double check when i get home

Grimothy:
@sparkison That was the part that reloads
  REVERB_APP_SECRET (and the other secrets) from the persisted config on boot. Without it, a container restart rotates the secret, so already-connected clients silently fail the Reverb auth
  check — that's why DVR status stops updating until you force-close and reopen the app.

  I can put up a follow-up PR to re-add just that piece if you want.

Issue #1329, opened 2026-07-26
Authenticate and replay-protect all m3u-proxy callbacks


Serph91P:
## Problem

Proxy failover, webhook, broadcast, and DVR callback paths do not share one fail-closed authentication contract. DVR callbacks accept all requests when no token is configured.

## Scope

- Require authentication before every callback side effect.
- Use HMAC bound to method, path, payload digest, timestamp, and nonce or event ID.
- Reject missing, invalid, expired, and replayed requests.
- Throttle callbacks and retain only an explicit local-development bypass.

## Acceptance criteria

- All callback routes use the same reviewed verifier contract.
- Valid retries remain idempotent.
- Feature tests prove rejection occurs before state changes.

## Triage

- Type: Bug
- Priority: P0
- Size: M
- Estimate: 5

sparkison:
Resolved in #1364

Issue #1328, opened 2026-07-26
Enforce PlaylistAuth ownership and quotas for all Xtream DVR actions


Serph91P:
## Problem

Experimental DVR capability advertising checks per-credential access, but action dispatch and stream access do not consistently repeat that scope.

## Scope

- Centralize authorization for every DVR action.
- Deny disabled or expired credentials.
- Stamp rules and recordings with playlist_auth_id.
- Scope list, detail, cancel, delete, and playback while preserving explicit owner behavior.
- Enforce per-credential concurrent and storage quotas.

## Acceptance criteria

- Two credentials on one effective playlist cannot observe or mutate each other.
- Stable error codes cover denial and quota exhaustion.
- Regular, custom, merged, and alias playlist tests pass.

## Triage

- Type: Bug
- Priority: P0
- Size: L
- Estimate: 8

sparkison:
Resolved in #1364

Issue #1331, opened 2026-07-26
Require signed authorization for media-server proxy streams


Serph91P:
## Problem

Media-server proxy routes accept integration and item IDs without binding access to a playlist, viewer, or credential.

## Scope

- Require a short-lived signed ticket or authenticated TV session.
- Bind integration, item, requester identity, operation, and expiry.
- Validate ordinary, local, and WebDAV paths consistently.

## Acceptance criteria

- Missing, foreign, expired, and modified tickets fail closed.
- Tickets cannot cross integrations or items.
- Range, redirect, and transport tests pass.

## Triage

- Type: Bug
- Priority: P0
- Size: M
- Estimate: 5

sparkison:
Resolved in #1364

Issue #1305, opened 2026-07-19
Add authenticated, client-scoped proxy stream cleanup for m3u-tv


Serph91P:
## Problem

`m3u-tv` has a server-transcode recovery abstraction, but a production client cannot complete the proxy stream lifecycle safely against `m3u-editor`.

Starting playback can reuse the existing authenticated Xtream stream routes with `proxy=true`, an explicitly selected `profile`, and a generated `client_id`. The current stop route at `POST /api/m3u-proxy/player-stream/stop` is intended for the browser player. It is outside the authenticated TV route group, accepts `id`, `type`, and `client_id`, and does not verify that the requested media belongs to the caller's playlist.

This blocks a secure implementation of m3ue/m3u-tv#110. The client must not call the browser stop route with user-controlled media IDs.

## Scope

Add an authenticated, client-scoped TV API operation that releases a proxy playback client without introducing a generic arbitrary-URL transcode endpoint.

Suggested route:

`POST /api/tv/{username}/{password}/player-stream/stop`

Suggested request:

```json
{
  "type": "live",
  "stream_id": 123,
  "client_id": "generated-playback-client-id"
}
```

Supported types should map to the existing proxy metadata:

- `live`, `vod`, and `catchup` use `channel_id`
- `series` uses `episode_id`

## Implementation plan

1. Reuse the same Xtream credential and playlist resolution rules used by the stream routes. Do not create a second, weaker authentication model.
2. Validate that the requested enabled channel or episode belongs to the authenticated playlist, merged playlist, custom playlist, alias, or allowed `PlaylistAuth` assignment before touching proxy state.
3. Validate `client_id` with a bounded length and a conservative character set.
4. Call `M3uProxyService::stopStreamsByMetadata` with `force: false` and the exact validated `client_id` so another viewer of the pooled stream remains connected.
5. Make repeated stops idempotent. Return `204` when the caller's client is already gone.
6. Return a clear authentication or authorization failure for invalid credentials and out-of-scope media. Do not reveal whether another playlist owns the supplied ID.
7. Keep the existing browser `sendBeacon` route for the web player unless it can be migrated separately without breaking browser teardown.
8. Do not log path credentials, raw upstream URLs, or proxy URLs containing credentials.

## HLS lifecycle contract

- The same generated `client_id` is appended to the authenticated stream URL before the initial request.
- A stop releases only that client registration, never the pooled stream unconditionally.
[...]

PR #1364, opened 2026-08-03
chore: Close unauthenticated access on media-proxy, EPG, and player-stream-stop routes

### LINKED ISSUE #1330: Authenticate EPG APIs and remove credential-bearing fallback playback URLs
## Problem

Experimental EPG APIs can return stream-capable URLs without consistently authenticating the requested playlist and may substitute owner identity when credentials are absent.

## Scope

- Require a valid session or playlist credential for private EPG responses.
- Never substitute owner credentials into responses.
- Separate optional public metadata from playable private output.
- Bound page size and date ranges.

## Acceptance criteria

- Unauthenticated and foreign requests cannot obtain playback URLs or credentials.
- Public mode is explicitly non-playable.
- Missing, invalid, expired, foreign, and oversized request tests pass.

## Triage

- Type: Bug
- Priority: P0
- Size: M
- Estimate: 5

sparkison:
Fixes #1330, #1331, #1305.

Three related gaps let unauthenticated or unverified callers reach stream-serving endpoints without proving ownership of the underlying playlist:

- EPG API (/api/epg/playlist/{uuid}/data) defaulted to the playlist owner's real Xtream username/password whenever the caller didn't supply credentials, embedding them in returned stream URLs. It now only returns playable URLs for an authenticated panel session or credentials that resolve to that exact playlist; otherwise the response is metadata-only (playable: false).
- Media-server proxy (`/media-server/*`, `/local-media/*`, `/webdav-media/*`) accepted any integration/item ID with no auth at all. These routes now require a signed, 24-hour-limited URL (ValidateSignature); only the app's own URL generation can produce a valid link.
- player-stream/stop (`/api/m3u-proxy/player-stream/stop`) let any caller stop any other viewer's stream by ID, with no auth or ownership check. It's now gated by the panel session and the existing Channel/Episode policies. A new authenticated POST /api/tv/{username}/{password}/player-stream/stop endpoint was also added for the m3u-tv Flutter client, scoped to its Xtream/PlaylistAuth credential (see m3ue/m3u-tv#110).
All three fixes fail closed and are idempotent/non-oracle: denied and successful stop requests return the same response, so a caller can't use it to probe whether an ID belongs to someone else.

### DIFF
--- a/app/Http/Controllers/AIOStreamsProxyController.php
+++ b/app/Http/Controllers/AIOStreamsProxyController.php
@@ -82,7 +82,32 @@ class AIOStreamsProxyController extends Controller
             return response()->json(['error' => 'Failed to fetch streams from AIOStreams'], 502);
         }
 
-        return response()->json($response->json());
+        $data = $response->json();
+
+        // Never hand a raw resolved URL (often carrying the debrid account's own
+        // auth token) back to the caller — proxy every candidate the same way the
+        // browse UI and synced Channels/Episodes do. There's no durable row behind
+        // this call, so it goes through the short-lived cache-token "live" proxy.
+        // A stream list can hold dozens of candidates, so this is batched into one
+        // cache write (see generateAioStreamsLiveProxyUrls()) rather than one per
+        // candidate.
+        if (is_array($data['streams'] ?? null)) {
+            $rawUrlsByIndex = [];
+
+            foreach ($data['streams'] as $index => $stream) {
+                if (is_string($stream['url'] ?? null) && $stream['url'] !== '') {
+                    $rawUrlsByIndex[$index] = $stream['url'];
+                }
+            }
+
+            $proxiedUrlsByIndex = MediaServerProxyController::generateAioStreamsLiveProxyUrls($integrationId, $rawUrlsByIndex);
+
+            foreach ($proxiedUrlsByIndex as $index => $proxiedUrl) {
+                $data['streams'][$index]['url'] = $proxiedUrl;
+            }
+        }
+
+        return response()->json($data);
     }
 
     /**
--- a/app/Http/Controllers/Api/EpgApiController.php
+++ b/app/Http/Controllers/Api/EpgApiController.php
@@ -266,14 +266,15 @@ class EpgApiController extends Controller
         $requestedPassword = $request->get('password', null);
 
         // Only embed playable stream URLs for callers we can actually verify: an
-        // authenticated panel session (the in-app EPG viewer), or credentials that
-        // resolve to this exact playlist. Never substitute the playlist owner's
-        // real credentials for a caller we can't authenticate — unauthenticated
-        // callers get metadata only (see $isPlayable below).
+        // authenticated panel session that owns this specific playlist (the
+        // in-app EPG viewer), or credentials that resolve to this exact
+        // playlist. Never substitute the playlist owner's real credentials for
+        // a caller we can't authenticate as that owner — everyone else gets
+        // metadata only (see $isPlayable below).
         $username = null;
         $password = null;
 
-        if (Auth::check()) {
+        if (Auth::check() && Auth::id() === $playlist->user_id) {
             $username = $user->name ?? 'admin';
             $password = $playlist->uuid;
         } elseif ($requestedUsername && $requestedPassword) {
--- a/app/Http/Controllers/Api/EpgApiController.php
+++ b/app/Http/Controllers/Api/EpgApiController.php
@@ -18,6 +18,7 @@ use Carbon\Carbon;
 use Exception;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\URL;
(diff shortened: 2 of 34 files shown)


sparkison:
Fixes #1329.

Follow-up: fixed test regressions from Phase 1/2 security fixes

Running the full suite surfaced 33 failures caused by the auth changes in this PR. All are now fixed (99/99 targeted tests passing, Pint clean, `composer security:scan` clean). Root causes and fixes:

### 1. Phase 1's fail-closed callback gate broke pre-existing broadcast/failover tests
`NetworkBroadcastFailureHandlingTest`, `NetworkBroadcastTransitionTest`, `ResolveEpisodeFailoverUrlTest`, and two cases in `ProxyLogRedactionTest` hit `/api/m3u-proxy/broadcast/callback` and `/api/m3u-proxy/failover-resolver` without a token — previously fine under the old fail-open default, now correctly rejected with 401.

**Fix:** these suites test broadcast/failover *business logic*, not the auth gate itself (that's covered separately by `M3uProxyCallbackAuthTest`), so:
- Added `config(['proxy.m3u_proxy_token' => null, 'proxy.allow_unauthenticated_callbacks' => true])` to their `beforeEach` blocks — the explicit local-dev bypass this feature was built for.
- Note: `m3u_proxy_token` had to be explicitly nulled, not just left unset, because a real token is present in this environment's own `.env` and was silently defeating the "no token configured" bypass path.
- `ProxyLogRedactionTest` already configures its own token, so its two requests just got an `X-API-Token` header added instead.

### 2. `stopPlayerStream` conflated two different concerns
My rewrite folded "malformed request" (missing `id`/`type`) into the same silent-204 response designed to hide ownership info from probing. That's wrong — structural validation isn't an oracle risk and should still 422.

**Fix:** restored 422 for missing/invalid `id`/`type`; the silent-204-regardless-of-outcome contract now applies only to the actual ownership/auth check. Updated `StopPlayerStreamTest` to authenticate as the real channel/episode owner (via new `actingAsChannelOwner()`/`actingAsEpisodeOwner()` helpers) so the id/type/client_id → proxy-request plumbing is still exercised through the new ownership gate, and fixed my own new test's now-incorrect 204 expectation to 422.

### 3. `LocalMediaStreamSecurityTest` built unsigned URLs
Its `streamUrl()` helper called `route('local-media.stream', [...])` directly, which Phase 1's signed-URL requirement now rejects (403) before the controller's own path-traversal checks ever run.

**Fix:** routed the helper through `MediaServerProxyController::generateLocalMediaStreamUrl()` — the same generator production code uses — so the suite validates path-traversal protection through a realistically signed URL instead of a bypassed one.

sparkison:
Fixes #1328.

Follow-up: DVR ownership enforcement 

Threads `PlaylistAuth` ownership through Xtream DVR dispatch, closing the gap where dispatch only scoped by `dvr_setting_id` (the shared playlist) and never checked which individual credential owned a rule/recording. Two credentials sharing one playlist could previously view, cancel, or delete each other's DVR recordings.

The schema was already in place — `playlist_auth_id` on `dvr_recordings`/`dvr_recording_rules` and the `playlistAuth()` relations landed in an earlier migration, and `DvrSchedulerService` already propagated `rule->playlist_auth_id` onto materialized recordings. The gap was entirely in `XtreamApiController`'s dispatch.

### Changes

1. **Dispatch-level gate**: the `DVR_ACTIONS` branch now rejects `playlist_auth`-authenticated callers whose credential has `dvr_enabled = false` — the same rule `canAdvertiseDvrFeature` already applied to feature *advertisement*, now applied to actual dispatch too. Previously a credential the TV app never showed DVR controls for could still call the DVR actions directly by hitting the endpoint.
2. **Ownership stamping**: `scheduleDvr` and `createDvrSeriesRule` stamp the created rule with `playlist_auth_id` from the requesting credential (`null` for the playlist owner).
3. **Ownership enforcement**: `getDvrRecordings`, `getDvrRecording`, `cancelDvrRecording`, and `deleteDvrRecording` filter by `playlist_auth_id` when the caller is a `PlaylistAuth` credential. A cross-credential lookup now 404s rather than exposing or mutating another viewer's recording.
4. **Owner exception preserved**: the playlist owner (`owner_auth`/`alias_auth`) is exempt from the filter and keeps full visibility/control across every credential's recordings — matches the "no regressions" convention for existing behavioral defaults.

**Not implemented (deliberately, scope):** per-credential concurrent/storage quota checks from the original proposal's step 4 — no issue calls for it and it's new product behavior, not an ownership-boundary fix. Flagging as a possible follow-on, not bundling it into a security fix.

### Tests
- New: `tests/Feature/XtreamDvrOwnershipTest.php` — cross-credential isolation on all four dispatch actions, owner override retaining full visibility, ownership stamping on rule creation, and the new `dvr_enabled` dispatch gate (parameterized across all six DVR actions).
- Updated: `tests/Feature/XtreamScheduleDvrTest.php` and `tests/Feature/XtreamCancelDeleteDvrTest.php` — their fixture credentials previously worked only because the `dvr_enabled` gate didn't exist yet; now explicitly set `dvr_enabled: true` and stamp `playlist_auth_id` on recordings they create, plus two new assertions locking in the stamping behavior.

sparkison:
...just realizing signed routes aren't going to work with the sync, they're going to go stale unless the media server is synced every 24hr or less.

I'm going to rethink this a bit and see if maybe we can retain the same URL's for the media browser integrations for authenticated users (`auth()->check()` true) and any subscribed users would need to use their own Xtream API generated URL to access. 

Not as secure as signed routes, but it's going to be confusing when things start returning `401` without any reason, at least this way auth users can update their passwords, or be removed, to prevent unauthorized access (same as every other Xtream API route).

PR #1371, opened 2026-08-07
feat(dvr): series-rule Xtream actions, options on create, and update endpoint

Grimothy:
Server side of series recording from EPG show search. Companion to **m3ue/m3u-tv#180**, which is blocked on this — the client calls these actions and is non-functional without them. This should land first.

## New Xtream actions

| action | purpose |
|---|---|
| `search_epg_shows` | Groups upcoming `EpgProgramme` rows by `SeriesKey`-normalized title across the playlist's EPG-mapped channels. Returns per-show channels, episode count, next airing, recent episodes, and `has_series_rule` / `series_rule_id`. |
| `list_dvr_series_rules` | Enabled Series rules for the playlist, scoped by `playlist_auth` where applicable. |
| `update_dvr_series_rule` | In-place edit of an existing rule. |
| `delete_dvr_series_rule` | Removes a rule by id. |

## `create_dvr_series_rule` extended

Now accepts `priority`, `start_early_seconds` and `end_late_seconds`, and **`channel_id` becomes optional**.

Omitting `channel_id` means *any channel*. `DvrSchedulerService::resolveSeriesEpgScope()` already scopes a rule to every EPG-mapped channel in the playlist when both `channel_id` and `source_channel_id` are null, so neither is set. Note this deliberately differs from the Browse Shows page, whose "any channel" option (`seriesChannelId === 0`) sets `source_channel_id` and therefore *narrows* scope to the browsing channel.

**Omit-to-inherit** is applied consistently: a param that is absent, blank or non-numeric is left out of the create entirely so the `DvrSetting` default applies. Blank is never coerced to `0` — for the padding fields `0` means *no padding*, which is not the same as *inherit*. The same treatment was applied to the pre-existing `keep_last` handling, which turned a blank value into `0`.

## Fix: `series_mode=new_flag` was silently stored as `all`

`DvrRecordingRule::saving` carries a legacy `new_only` → `series_mode` migration:

```php
} elseif (! $rawNewOnly && $rule->series_mode === DvrSeriesMode::NewFlag) {
    $rule->series_mode = DvrSeriesMode::All;
}
```

`new_only` defaults to `false` and the Xtream controller never set it, so **every** rule created through the API with `series_mode=new_flag` was rewritten to `all` on save — "New episodes only" silently became "All episodes". The Browse Shows page escapes this because `recordSeriesWithOptions()` passes both fields.

Fixed by setting `'new_only' => ($seriesMode === DvrSeriesMode::NewFlag)` on create and update. It has to be exactly that equality: setting `new_only = true` alongside `unique_se` trips the hook's *first* branch and clobbers the mode to `new_flag`. **The shared saving hook is deliberately left untouched**, since it is also used by the Filament resources.

Verified across all 12 `series_mode` × `match_mode` combinations against a local container — `new_flag` was the only broken value, and all four match modes always round-tripped correctly.

## Duplicate guard

Creating a second Series rule for the same show now returns **409** with the existing `rule_id` and `duplicate: true`, so a client can show "already recording" instead of a generic failure. Previously the API would happily create unlimited duplicate rules for one show, while the Browse Shows page refused them.

The lookup uses the auto-derived, indexed `normalized_title` column via `SeriesKey::normalize` — matching what the model actually stores and what `search_epg_shows` groups by, rather than `EpgProgrammeNormalizer::cleanForSearch` as the web page uses. Worth noting those two normalizers coexist in the codebase and can disagree.

The guard is intentionally scoped by DVR setting + type + normalized title + auth, and **not** by channel, matching `GuestBrowseShows::createSeriesRule` — a second rule for the same show on a different channel is treated as a duplicate.

## Update semantics

Only fields present in the request are applied, so editing one setting never nulls out unrelated ones. `channel_id` present-but-blank switches a rule to any-channel; absent leaves it unchanged.

Edit is a genuine in-place update rather than delete-and-recreate: deleting cascades to the rule's recordings, which would destroy recording history and change the rule id.

`DvrSchedulerTick` is **not** dispatched — `DvrRecordingRule::boot()`'s `created` hook already calls `DvrSchedulerService::scheduleRuleImmediately()` for enabled rules, so the API path is covered.

## Testing

`tests/Feature/XtreamDvrOwnershipTest.php` — 18 passing / 61 assertions, Pint clean. Covers the new-flag lockstep in both directions (including that `new_only` must not rewrite a `unique_se` rule), present-fields-only updates, blank-channel → any-channel, and cross-credential 404s.

Also exercised end to end against a running container: the full 12-combination matrix, explicit `start_early_seconds=0` persisting as `0` rather than null, blank inheriting, non-numeric `priority` falling back to the column default, the 409 path, and `get_dvr_storage` from #1360 still responding alongside the new actions.

## Notes

- `search_epg_shows` returns `has_series_rule` and `series_rule_id` so a client can render correct toggle state without a second round trip.
- `priority` was added to `formatDvrSeriesRule`, which previously omitted it.
- Also included: the playlist title search in `search_epg_shows` uses `ILIKE` so lowercase queries match capitalised titles (PostgreSQL-specific).
- Per-rule `enable_comskip` is out of scope; the Browse Shows options form does not expose it either.

### DIFF
--- a/app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource.php
+++ b/app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource.php
@@ -166,7 +166,10 @@ class GuestDvrRuleResource extends Resource
                         ->when($record !== null, fn ($col) => $col->prepend(__('From Original Source'), 0))
                     : [])
                 ->searchable()
-                ->nullable(),
+                ->nullable()
+                ->helperText(fn (Get $get): ?string => self::isRuleType($get('type'), DvrRuleType::Series)
+                    ? __('Series rules only match against channels with EPG data mapped. If a rule never records, confirm this channel has an EPG source assigned.')
+                    : null),
 
             TextInput::make('series_title')
                 ->label(__('Series Title'))
--- a/app/Filament/Resources/DvrRecordingRules/DvrRecordingRuleResource.php
+++ b/app/Filament/Resources/DvrRecordingRules/DvrRecordingRuleResource.php
@@ -119,7 +119,10 @@ class DvrRecordingRuleResource extends Resource
                         ->pluck('title', 'id')
                         ->prepend(__('From Original Source'), 0))
                     ->searchable()
-                    ->nullable(),
+                    ->nullable()
+                    ->helperText(fn (Get $get): ?string => self::isRuleType($get('type'), DvrRuleType::Series)
+                        ? __('Series rules only match against channels with EPG data mapped. If a rule never records, confirm this channel has an EPG source assigned.')
+                        : null),
 
                 TextInput::make('series_title')
                     ->label(__('Series Title'))
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -18,6 +18,7 @@ use App\Models\CustomPlaylist;
 use App\Models\DvrRecording;
 use App\Models\DvrRecordingRule;
 use App\Models\Epg;
+use App\Models\EpgProgramme;
 use App\Models\Group;
 use App\Models\MergedPlaylist;
 use App\Models\Network;
@@ -37,6 +38,7 @@ use App\Services\EpgCacheService;
 use App\Services\LogoCacheService;
 use App\Services\M3uProxyService;
 use App\Settings\GeneralSettings;
+use App\Support\SeriesKey;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Request;
@@ -68,6 +70,9 @@ class XtreamApiController extends Controller
         'create_dvr_series_rule',
         'cancel_dvr_recording',
         'delete_dvr_recording',
+        'list_dvr_series_rules',
+        'delete_dvr_series_rule',
+        'search_epg_shows',
     ];
 
     private const FAVORITE_COLUMNS = [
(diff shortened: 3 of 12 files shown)


Serph91P (commented):
## Review: Xtream DVR ownership + series rules

Overall this is a solid, well-scoped backend change. The ownership model is consistently applied and the test coverage for credential-scoped DVR actions is good.

###  What looks good
- Ownership/scoping is enforced across all DVR endpoints; the playlist owner override works as intended.
- `tests/Feature/XtreamDvrOwnershipTest.php` covers the key security paths (23 tests, 76 assertions, all passing locally).
- Migrations for `dvr_recordings` and `dvr_recording_rules` are clean with appropriate indexes and constraints.
- Pint formatting passes for the touched files.

###  Discussion points / suggestions
1. **Wildcard behavior**: I verified that `title = "%"` matches and schedules unrelated shows across any channel. This is technically correct for `DvrMatchMode::Wildcard`, but a bare `%` rule is dangerous for end users. Consider rejecting or explicitly documenting it.
2. **Deleting a series rule leaves recordings in place**: Because `dvr_recordings.dvr_recording_rule_id` is set to `null` on delete, planned recordings remain active. This is a reasonable design, but the client should communicate it clearly.
3. **Naming consistency for `series_mode`**: Internally `DvrSeriesMode::NewOnly` is stored as `new_flag`, while the client-facing parameter seems to use `new_only` in places. Please double-check that the API contract with `m3u-tv#180` is identical on both sides.
4. **EPG scope**: A series rule is skipped when the channel has no `epg_channel_id`. This is correct, but relies on users keeping EPG mappings up to date.

### Local verification
- `vendor/bin/pest tests/Feature/XtreamDvrOwnershipTest.php` → 23 warnings, 76 assertions, no failures.
- Added two contract checks locally: deleting a series rule keeps existing recordings; `title="%"` schedules multiple shows. Both passed.

### Recommendation
Approve once the `new_only`/`new_flag` naming contract with the Flutter client is confirmed.

PR #1378, opened 2026-08-08
fix(push-device-tokens): format pairing code input as XXXX-XXXX

Grimothy:
## Summary
On the device pairing page, the code input showed a placeholder hint of `XXXX-XXXX` but never live-formatted as the user typed — they could type `xxxxxxxx` with no dash and it would just submit as-is. Cosmetic-only fix: the underlying value/validation/storage was never touched.

```php
TextInput::make('user_code')
    ->label(__('Device Code'))
    ->required()
    ->maxLength(12)
    ->placeholder('XXXX-XXXX')
    ->mask('9999-9999')
    ->stripCharacters('-'),
```

- `->mask('9999-9999')` gives live visual formatting (`XKQP-9F3T`) as the user types, via Filament's real `x-mask` Alpine directive.
- `->stripCharacters('-')` is a Filament state-cast that strips the dash from the bound Livewire state on the way in/out — the actual submitted value never contains a dash, so `normalizeUserCode()`, `approve()`, and existing validation are completely unaffected.

Nothing else changed.

## Test plan
- [x] `tests/Feature/DevicePairingPageTest.php` — 12/12 pass unchanged, including the existing dash-free/lowercase/whitespace-variant cases.
- [x] `vendor/bin/pint` clean.

### DIFF
--- a/app/Filament/Resources/PushDeviceTokens/Pages/ListPushDeviceTokens.php
+++ b/app/Filament/Resources/PushDeviceTokens/Pages/ListPushDeviceTokens.php
@@ -102,7 +102,7 @@ class ListPushDeviceTokens extends ListRecords
                                 ->required()
                                 ->maxLength(12)
                                 ->placeholder('XXXX-XXXX')
-                                ->mask('9999-9999')
+                                ->mask('****-****')
                                 ->stripCharacters('-'),
                             Select::make('playlist_auth_id')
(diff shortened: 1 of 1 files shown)


Grimothy:
This one doesnt matter ---- I was testing a new workflow :D

sparkison:
haha! Nice addition though 😄

Issue #1340, opened 2026-07-27
Disabling Categories/Groups does nothing?


singular918:
## Describe the bug

I would like to disable a bunch of categories/groups because they are of a language/country that I do not want.

## To Reproduce

1. Go to Series -> Categories.
2. Select a couple of Categories
3. Select "Bulk Action" -> Disable Categories

## Expected behavior

The Categories show some indication of being disabled and then are not included in the playlist output. 

## Actual behavior

Nothing happens.

Serph91P:
> ## Describe the bug
> 
> I would like to disable a bunch of categories/groups because they are of a language/country that I do not want.
> ## To Reproduce
> 
> Steps to reproduce the behavior:
> 
>     1. Go to Series -> Categories.
> 
>     2. Select a couple of Categories
> 
>     3. Select "Bulk Action" -> Disable Categories
> 
> 
> ## Expected behavior
> 
> The Categories show some indication of being disabled and then are not included in the playlist output.
> ## Actual behavior
> 
> Nothing happens.

You can either in the pre processing of the playlist dont sync the categories you don't want or disable the categories after you have synced the playlist. But you have to differentiate there are two options. The on disabled auto sync for the categories the other actually disables all series inside the categories. 

<img width="450" height="513" alt="Image" src="https://github.com/user-attachments/assets/e3fa8ed5-b284-4cbd-b49a-456348904567" />

singular918:
Not sure if I understand fully. Disabling them in pre-processing is hard because you have specify prefixes, patterns, categories names, etc and the playlist from provider is all over the place. I want to look at them one by one and disable them while the new ones (that the provider may add in future) may continue to be created and synced.

Post-Processing - Yeah, I do see two options. "Disable Category Series" makes the count of the series inside the category zero but the empty category is still there synced to the IPTV Player right? I am trying to not see it at all in the Player. 

"Disable Categories" - What does this do? Are you saying this just disables future auto-sync for this category with the provider? if thats the case then shouldn't there be an indication/icon that auto-sync is disabled and give an option to enable it if someone wanted to?

Anyway, I am coming form iptveditor world and that allows you to disable categories which changes color in the interface to indicate that these are disabled. These still sync with the provider, you can search in them to see if there is something you wanted and can enable them back if needed. They don't output in the IPTV Player playlist. I was looking for something like that but I guess that is not possible? 

How are most people here hiding the categories that they are not intrested in?

Serph91P:
[...]

PR #1401, opened 2026-08-11
fix: hide disabled series categories from Xtream output

gemini-it:
## Summary

- exclude disabled regular series categories from get_series_categories
- exclude their series from get_series and direct get_series_info access
- keep custom playlist tag behavior unchanged
- keep uncategorized series visible for backward compatibility

## Why

The editor records Category.enabled, but the Xtream series endpoints currently filter only Series.enabled. A disabled language or country category therefore remains visible to Xtream clients. This is the remaining output-side behavior described in #1340.

## Tests

- enabled categories and series remain visible
- disabled categories, listings, and direct series info are hidden
- uncategorized series remain visible
- existing Xtream controller coverage remains green

Pint passes and all 38 scoped tests pass.

### DIFF
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -992,6 +992,13 @@ class XtreamApiController extends Controller
                 ->orderBy('series.sort', 'asc')
                 ->with(['tags', 'category']);
 
+            if (! $isCustomPlaylist) {
+                $seriesQuery->where(function ($query) {
+                    $query->whereNull('series.category_id')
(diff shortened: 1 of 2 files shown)


sparkison:
The enable you're speaking of is actually the "Auto enable" feature used to automatically enabled any new channels that appear within the group or category on sync. So you'd enable "US" group to have all new us grouped channels automatically enabled for example, it was not intended to enable/disable the group or category itself.

We'd need the PR to include an actual enable/disable at the group/category level to implement this, which doesn't exist currently. Closing for now.

PR #1402, opened 2026-08-11
fix: map Xtream season metadata by season number

gemini-it:
## Summary

- resolve season metadata using the explicit provider season_number instead of a JSON list offset
- retain compatibility with associative and zero-indexed provider responses
- update season names when metadata is refreshed
- update episode season_id during upsert so provider corrections can re-parent an existing episode

## Why

Xtream commonly returns seasons as a zero-indexed list while episodes are keyed by actual season numbers. Indexing the list with the episode season key shifts season metadata by one and can produce duplicated or mislabeled seasons.

## Tests

- maps two zero-indexed provider seasons to season 1 and season 2 correctly
- moves an existing episode to the corrected season on refresh

Pint passes and both scoped regression tests pass.

### DIFF
--- a/app/Models/Season.php
+++ b/app/Models/Season.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Casts\Attribute;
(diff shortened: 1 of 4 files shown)


sparkison:
thank you! A couple minor tweaks, but looks good otherwise 🙏

Issue #1395, opened 2026-08-10
DVR: VOD integration can silently skip a completed recording with no surfaced signal


Grimothy:
## Is your feature request related to a problem? Please describe.

A DVR recording can complete perfectly — correct file on disk, `status=completed`, TMDB show *and* episode metadata enriched, `season`/`episode` populated — and still never appear in the library, with **no signal anywhere in the UI** that anything went wrong.

That happened in practice: a recording of "The Real Housewives: Ultimate Girls Trip" S5E1 finished cleanly (3.2 GB, 5490 s, `error_message: null`, `post_processing_step: null`) but no `series` and no `episodes` row was ever created. The DVR Recordings screen showed it as a healthy completed recording. The only trace of the failure was a single line in the Laravel log:

```
DvrVodIntegration: missing playlist_id or user_id for recording 96 — skipping
```

Diagnosing it took a full troubleshooting session and a series of ad-hoc DB queries, purely because there was nothing to look at. The specific root cause is being fixed separately (`DvrVodIntegrationService::integrateRecording()` only reads `$setting->playlist_id`, so a DVR setting owned by a merged or custom playlist bails out). This issue is about the *class* of problem, not that one cause: **VOD integration can fail or be skipped for any reason and the outcome is never persisted or surfaced.**

Concretely, `integrateRecording()` has three early-return paths (no `DvrSetting`, missing `playlist_id`/`user_id`, plus a caught-and-rethrown exception path) and every one of them ends in `Log::warning` / `Log::error` only. There is no `integration_status`-style column on `dvr_recordings` and nothing on the model records whether integration ran, was skipped, or threw.

## Describe the solution you'd like

Persist and surface the integration outcome so a skipped or failed integration is visible without reading logs:

1. **Persist the outcome** on `dvr_recordings` — e.g. an integration status (`pending` / `integrated` / `skipped` / `failed`) plus a short reason string, written by `IntegrateDvrRecordingToVod` / `DvrVodIntegrationService` on each of the early-return and exception paths.
2. **Surface it on the Recordings screen** — a completed recording that never made it into the library should be visually distinguishable from one that did. The relations to check already exist (`DvrRecording::vodChannel()` / `vodEpisode()`), so even without a new column the *absence* of both is detectable today for a recording that should have produced one.
[...]

PR #1404, opened 2026-08-11
fix(dvr): re-broadcast status once VOD integration actually finishes

### LINKED ISSUE #1403: DVR: completion broadcast fires before VOD/Series integration finishes, so clients refresh too early
## Describe the bug

A DVR recording finishes, `status` flips to `completed`, and the m3u-tv client (which listens for exactly that transition to refresh its Movies/Series lists — see `m3ue/m3u-tv` PR [#200](https://github.com/m3ue/m3u-tv/pull/200)) refreshes immediately and finds nothing new. The recording is not lost — it's genuinely in the library once you manually refresh later — but the automatic refresh this was built for doesn't do anything, because the library doesn't exist yet at the moment the client is told to look.

## Root cause

`DvrPostProcessorService::process()` does the file-side work and the library-side work in the wrong order relative to when it announces "done":

1. **Step 3** (`app/Services/DvrPostProcessorService.php:177-206`) dispatches the async chain that actually builds the library entry — either `EnrichDvrMetadata::dispatch()` (which itself dispatches `IntegrateDvrRecordingToVod` only after a TMDB/TVMaze lookup completes, `app/Jobs/EnrichDvrMetadata.php:68`) or `IntegrateDvrRecordingToVod::dispatch()` directly when enrichment is disabled. Either way, this just enqueues work on the `dvr-post` queue — it does not wait for it.
2. **Step 4**, immediately after (`app/Services/DvrPostProcessorService.php:231-235`), calls `$recording->update(['status' => Completed->value, ...])`.
3. `DvrRecording`'s `static::updated` hook (`app/Models/DvrRecording.php:64-66`) fires `broadcastStatus()` automatically whenever `status` changes — so the completion push goes out **the instant step 4's `update()` lands**, before the job queued in step 3 has necessarily run.

So the client is told "completed" — and reacts by refreshing VOD/Series — before `IntegrateDvrRecordingToVod::handle()` (`app/Jobs/IntegrateDvrRecordingToVod.php`) has created the `Series`/`Episode`/`Channel` rows that refresh is supposed to surface. The window is a queue-latency gap in the simple case, and widens further when metadata enrichment is enabled, since that adds a synchronous TMDB/TVMaze network round-trip before `IntegrateDvrRecordingToVod` is even dispatched.

No second push ever arrives once the job actually finishes — `status` doesn't change again — so nothing re-triggers the client on its own. Only a manual user-initiated refresh, issued after the queue has had time to catch up, finds the content.

This is a distinct failure mode from #1395/#1396 — those cover integration silently *skipping* or *failing*. Here integration succeeds correctly; it just finishes **after** the client was already told it was done.

## Reproduction

1. Enable metadata enrichment on a DVR setting (or otherwise ensure `dvr-post`/`dvr-meta` queue workers have any latency relative to the request that flips `status`).
2. Let a recording complete.
3. Watch a connected client that reacts to the `completed` push (e.g. m3u-tv) — Movies/Series will not show the new entry.
4. Manually pull-to-refresh (or restart) the client after the queue has caught up — the entry is there.

## Proposed fix

Re-broadcast the recording's status a second time from the *end* of the actual integration job, once the library rows genuinely exist:

```php
// app/Jobs/IntegrateDvrRecordingToVod.php, end of handle()
$recording->update(['post_processing_step' => null]);

$recording->broadcastStatus();   // NEW — re-announce now that the library actually reflects this recording
```

`broadcastStatus()` (`app/Models/DvrRecording.php:298-303`) is a plain, side-effect-free rebuild-and-broadcast of the current state — safe to call a second time with an unchanged `status` value. On the client side this needs **no change at all**: m3u-tv's push handler (`_onDvrStatusPush` → `_refreshDvrRecordingDetail`, `flutter_client/lib/services/app_state_controller.dart:1106-1196`) re-fetches and re-classifies unconditionally on every push it receives — it does not compare against a previously-seen status — so a second `completed` push naturally triggers exactly the refresh that was missing, this time with real data behind it.

This interacts correctly with existing failure semantics without extra work:
- If `integrateRecording()` throws, `DvrVodIntegrationService.php`'s catch block rethrows and `handle()` never reaches the new line — no spurious second broadcast on genuine failure, and the job's existing `tries = 3` retry still applies.
- If `integrateRecording()` early-returns (the #1396/#1395 skip paths — no resolvable playlist, no user, etc.), `handle()` still reaches the new line and fires a redundant no-op broadcast. This is a deliberate simplification, not an oversight: the client-side refresh this triggers is already designed to be safe to run with nothing new to find (`_classifyContentTargets`'s own doc comment: "the issue permits over-refresh, never under-refresh" — same principle from m3u-tv #179/#200 applies here).

## Alternatives considered

- **Delay the `Completed` status flip itself until integration finishes.** Rejected: `status = completed` has other legitimate consumers that are correctly ready *before* library integration — the file is already playable via the "Watch" action, and the completed/failed `TvNotification` (`DvrRecording::notifyTv`) is meant to fire promptly. Moving the status flip would delay those for no reason.
- **A new, distinct broadcast event for "library updated."** Rejected for now: correct in principle, but requires a coordinated client change to listen for a second event type, where re-using the existing status broadcast gets the fix out with zero client work and no new wire contract.
- **Gate the second broadcast on whether integration actually created/updated a row.** Would avoid the redundant no-op broadcast on the skip path, but `integrateRecording()` currently returns `void`; plumbing a bool through is a larger change for a marginal benefit given the "over-refresh is fine" precedent already established.

## Test plan (for whoever picks this up)

- [ ] `IntegrateDvrRecordingToVod` test: mock/spy `DvrRecording::broadcastStatus()` (or assert a `DvrRecordingStatusEvent` is broadcast) and confirm it fires once at the end of a successful `handle()`.
- [ ] Confirm it does **not** fire when `integrateRecording()` throws (job fails/retries as today).
- [ ] Confirm the early-return/skip paths still complete `handle()` without error (the redundant broadcast is harmless, per above).
- [ ] Manual/integration check: with metadata enrichment enabled, confirm a client receives two `completed` pushes for one recording — the second one after `Series`/`Episode` rows exist.

Related: #1395, #1396, m3u-tv#200 (m3ue/m3u-tv).

Grimothy:
Closes #1403

## Problem

A completed recording's status push goes out before the library rows it announces actually exist.

`DvrPostProcessorService::process()` dispatches the VOD integration job (`IntegrateDvrRecordingToVod`, either directly or via `EnrichDvrMetadata` when metadata enrichment is enabled) onto a queue — fire-and-forget, no wait — then immediately flips `status` to `Completed`. `DvrRecording`'s `updated` hook auto-broadcasts on any status change, so the push goes out the instant that happens.

A connected client (e.g. m3u-tv) refreshes its library on that push and finds nothing new, because the queued job hasn't created the `Series`/`Episode`/`Channel` rows yet. No second push ever arrives once it does, since `status` doesn't change again — only a manual client refresh, issued after the queue caught up, finds the content.

Full root-cause trace with file:line citations is in #1403.

## Fix

One line at the end of `IntegrateDvrRecordingToVod::handle()`:

```php
$recording->broadcastStatus();
```

Re-announces the same (unchanged) status once the library rows genuinely exist. `broadcastStatus()` is a plain rebuild-and-broadcast, safe to call a second time.

**No client-side change needed.** m3u-tv's push handler re-fetches and re-classifies unconditionally on every push it receives, with no comparison against a previously-seen status — so the second push naturally triggers the refresh the first one couldn't.

Deliberately not gated on whether integration actually did anything:
- The #1395/#1396 early-return paths (no resolvable playlist, no user) still reach this line and produce a redundant no-op broadcast. Matches the "over-refresh, never under-refresh" precedent already established client-side for this exact feature.
- A thrown exception in `integrateRecording()` means `handle()` never reaches this line at all — no spurious broadcast on genuine failure, the job's existing `tries = 3` retry is unaffected.

Alternatives considered and rejected (detailed in #1403): delaying the `Completed` status flip itself (breaks the "Watch" button and notification, which are legitimately ready earlier), and a new distinct event type for "library updated" (needs a coordinated client change for no benefit over reusing the existing push).

## Test plan

- 3 new cases in `tests/Feature/DvrVodIntegrationServiceTest.php`:
  - re-broadcasts exactly once on a successful `handle()`
  - does **not** re-broadcast when `integrateRecording()` throws — retry semantics preserved
  - completes `handle()` without error on the no-resolvable-playlist skip path. (`broadcastStatus()` itself no-ops there — `DvrRecordingStatusEvent`'s own `owner()` guard returns null with nothing to key a Reverb channel on, not new behaviour from this fix.)
- `vendor/bin/pint --dirty --test` → passed
- `vendor/bin/pest tests/Feature/DvrVodIntegrationServiceTest.php` → 47 passed (128 assertions) — 44 pre-existing + 3 new
- **Mutation-checked**: reverting the fix fails exactly the "fires once" case (`assertDispatchedTimes` 0 vs 1); the throw-path and skip-path cases are unaffected, as expected since neither reaches the fix's line either way. Restored, 47/47 green.
- Full `php artisan test`: not chased further. All 762 failures in this worktree's environment are `ConnectionException` against a Redis port this worktree's environment never started, spanning domains with no plausible interaction with this change (AIOStreams, EDL endpoints, Xtream network fallback, regex tester).

### DIFF
--- a/app/Jobs/IntegrateDvrRecordingToVod.php
+++ b/app/Jobs/IntegrateDvrRecordingToVod.php
@@ -50,6 +50,15 @@ class IntegrateDvrRecordingToVod implements ShouldQueue
         // Clear the step label now that the full pipeline is done
         $recording->update(['post_processing_step' => null]);
 
+        // Re-broadcast the current status so connected TV clients know the
+        // library entry now exists. The completion broadcast fired from
+        // DvrPostProcessorService goes out the instant status flips to
+        // Completed, *before* this queued job has a chance to create the
+        // Series/Episode/Channel rows — so without this second push, clients
+        // refreshed on the first push and saw nothing. broadcastStatus() is
+        // safe to call with an unchanged status (issue #1403).
+        $recording->broadcastStatus();
+
         Log::info("DVR VOD integration complete for recording {$recording->id}", [
             'recording_id' => $recording->id,
         ]);
--- a/tests/Feature/DvrVodIntegrationServiceTest.php
+++ b/tests/Feature/DvrVodIntegrationServiceTest.php
@@ -19,6 +19,7 @@
  */
 
 use App\Enums\DvrRecordingStatus;
+use App\Events\DvrRecordingStatusEvent;
 use App\Jobs\EnrichDvrMetadata;
 use App\Jobs\IntegrateDvrRecordingToVod;
 use App\Models\Channel;
@@ -37,6 +38,7 @@ use App\Services\DvrVodIntegrationService;
 use Carbon\Carbon;
 use Filament\Notifications\DatabaseNotification;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Notification;
 use Illuminate\Support\Facades\Queue;
 use Illuminate\Support\Facades\Storage;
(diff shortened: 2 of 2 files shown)


sparkison:
no issues! Merging 😉

PR #1405, opened 2026-08-11
fix(guest-panel): fail closed when getCurrentPlaylistAuth() resolves to null

### LINKED ISSUE #1398: Guest panel: guests can see other guests' DVR recordings and rules
## Describe the bug

On the guest panel, a guest can see DVR recordings and series rules that belong to **other guests** (and to the playlist owner). The queries backing those lists are scoped by `dvr_setting_id` only — there is no `playlist_auth_id` filter — so every guest sharing a playlist sees the whole DVR library: titles, channels, scheduled times, statuses.

The *actions* are correctly ownership-gated (the `cancel` action, and the play action added in #1366 both check `playlist_auth_id === $currentAuth->id`), so the symptom looks odd rather than obviously wrong: a guest sees N recordings but can only act on their own. That mismatch is what surfaced it.

`app/Filament/GuestPanel/Pages/GuestRequestContent.php:193` already does the right thing (`->where('playlist_auth_id', $auth->id)`), so the correct pattern exists in the codebase — these sites just don't use it.

## Expected behaviour

A guest should only see their own recordings and their own rules.

## Affected sites — and an important distinction

**These leak and should be scoped to the current guest:**

| File | Line | What it backs |
|---|---|---|
| `app/Filament/GuestPanel/Resources/DvrRecordings/GuestDvrRecordingResource.php` | 141 | the recordings list |
| `app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource.php` | 108 | the series-rules list |
| `app/Filament/GuestPanel/Widgets/GuestScheduledSeriesWidget.php` | 30 | the scheduled-series widget |

**These look like the same pattern but must NOT be scoped:**

| File | Line | Why |
|---|---|---|
| `app/Filament/GuestPanel/Pages/GuestBrowseShows.php` | 328 | Once-rule duplicate check |
| `app/Filament/GuestPanel/Pages/GuestBrowseShows.php` | 467 | Series-rule duplicate check |
| `app/Filament/GuestPanel/Pages/GuestBrowseShows.php` | 623 | builds the "already scheduled" indicators for the browse UI |

Lines 328 and 467 are **integrity** checks, not display. They exist to stop a second rule being created for the same programme on the same DVR setting. Scoping them per-guest would let guest A and guest B each create a Once rule for the same programme, and the server would then **record it twice**. A find-and-replace that adds `playlist_auth_id` everywhere would introduce that regression.

Line 623 has to match 328/467. It builds the "already scheduled" markers shown while browsing. If it were scoped but the duplicate checks were not, a guest would see a show as unscheduled, click record, and be rejected with "A Once rule for this programme already exists" — the UI and the guard would disagree. So it stays unscoped for consistency with the thing it is previewing.

So the rule is: **scope the display queries, leave the integrity checks and their matching UI indicators alone.**

## Open design question

Once the display queries are scoped, recordings created by the **playlist owner** (`playlist_auth_id IS NULL`) become invisible to every guest, since no guest's id will match null. That is probably correct — they aren't a guest's recordings — but it is a deliberate behaviour change worth confirming, because a shared/family DVR setup might reasonably expect owner-created recordings to be visible to everyone on the playlist.

`GuestDvrRecordingResource::canAccess()` requires a non-null `PlaylistAuth` (via `guestCanAccessDvr()`), so there is no "owner browsing the guest panel" path to preserve — but the scoped query should still **fail closed** (return an empty result) if the auth is somehow null, rather than falling back to unscoped.

## Additional context

Observed on a merged playlist: DVR setting owned by `merged_playlist_id`, with two completed recordings under it — one belonging to a guest, one created by the playlist owner. The guest saw both listed, with an action on only their own.

Related: #1366 / #1397 added the play action and its ownership gate. This issue is deliberately separate because it changes data visibility rather than adding a control, and it spans several files.

Grimothy:
Closes #1398

## Bug

The three guest-panel DVR query entry points all share the same pattern: they filter on `$currentAuth?->id`, where `$currentAuth` comes from `getCurrentPlaylistAuth()`. When that helper returns `null`, Laravel's query builder translates the `null` right-hand side into `WHERE playlist_auth_id IS NULL` — which doesn't fail closed, it returns the playlist owner's records to the guest.

Affected entry points:
- `app/Filament/GuestPanel/Resources/DvrRecordings/GuestDvrRecordingResource::getEloquentQuery()`
- `app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource::getEloquentQuery()`
- `app/Filament/GuestPanel/Widgets/GuestScheduledSeriesWidget::getSeriesRules()`

## Fix

Added `&& ! static::isOwnerAuth()` to the null-auth guard in each entry point. When `getCurrentPlaylistAuth()` returns `null` AND the request is NOT in the owner-auth path (no `PlaylistAuth` row), return an empty result set instead of leaking the owner's records.

`isOwnerAuth()` is already defined in `app/Filament/GuestPanel/Pages/Concerns/HasGuestDvr.php` and is the canonical predicate for the owner-auth path; reusing it preserves the existing two owner-auth tests that legitimately rely on `getCurrentPlaylistAuth()` returning `null` for the playlist owner.

## Tests

Added 7 regression tests across the three affected files. All assert that when `getCurrentPlaylistAuth()` resolves to `null` (stale / mismatched state), the entry point returns an empty collection rather than the playlist owner's records.

- `tests/Feature/GuestDvrRecordingResourceTest.php` — table query, header stat, nav badge (3 tests)
- `tests/Feature/GuestDvrRuleResourceTest.php` — table query, header stat (2 tests)
- `tests/Feature/GuestScheduledSeriesWidgetTest.php` — `getSeriesRules()`, nav badge (2 tests)

## Verification

- `vendor/bin/pest` on the three target files: 72 tests pass (86 assertions), up from 65 pre-fix
- `vendor/bin/pint --dirty --test`: clean
- Mutation check: 3 rounds, each round reverted exactly one of the three guards, each round failed only that file's new tests, the other 70 stayed green — confirms each guard is independently load-bearing

### DIFF
--- a/app/Filament/GuestPanel/Resources/DvrRecordings/GuestDvrRecordingResource.php
+++ b/app/Filament/GuestPanel/Resources/DvrRecordings/GuestDvrRecordingResource.php
@@ -168,6 +168,19 @@ class GuestDvrRecordingResource extends Resource
 
         $currentAuth = static::getCurrentPlaylistAuth();
 
+        // Stale session: the credentials no longer resolve to a live
+        // PlaylistAuth row (e.g. revoked/disabled mid-session). Without this
+        // guard, Laravel's query builder turns `where(col, null)` into
+        // `whereNull(col)` and would happily return the playlist OWNER's
+        // recordings (the only ones with playlist_auth_id = null), re-opening
+        // the leak issue #1398 exists to close. isOwnerAuth() must be allowed
+        // through — the owner has no PlaylistAuth row, so getCurrentPlaylistAuth()
+        // legitimately returns null for them, and they own the recordings with
+        // playlist_auth_id = null.
+        if (! $currentAuth && ! static::isOwnerAuth()) {
+            return parent::getEloquentQuery()->whereRaw('1 = 0');
+        }
+
         return parent::getEloquentQuery()
             ->with(['channel', 'playlistAuth', 'dvrSetting.playlist', 'dvrSetting.customPlaylist', 'dvrSetting.mergedPlaylist'])
             ->where('dvr_setting_id', $dvrSetting->id)
--- a/app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource.php
+++ b/app/Filament/GuestPanel/Resources/DvrRules/GuestDvrRuleResource.php
@@ -119,6 +119,19 @@ class GuestDvrRuleResource extends Resource
 
         $currentAuth = static::getCurrentPlaylistAuth();
 
+        // Stale session: the credentials no longer resolve to a live
+        // PlaylistAuth row (e.g. revoked/disabled mid-session). Without this
+        // guard, Laravel's query builder turns `where(col, null)` into
+        // `whereNull(col)` and would happily return the playlist OWNER's
+        // rules (the only ones with playlist_auth_id = null), re-opening the
+        // leak issue #1398 exists to close. isOwnerAuth() must be allowed
+        // through — the owner has no PlaylistAuth row, so getCurrentPlaylistAuth()
+        // legitimately returns null for them, and they own the rules with
+        // playlist_auth_id = null.
+        if (! $currentAuth && ! static::isOwnerAuth()) {
+            return parent::getEloquentQuery()->whereRaw('1 = 0');
+        }
+
         return parent::getEloquentQuery()
             ->with(['channel', 'playlistAuth'])
             ->where('dvr_setting_id', $dvrSetting->id)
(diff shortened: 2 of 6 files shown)


sparkison:
Already addressed here e967ce4913e24f4c062b092f2b4322fe186c32b5 here 658fe5a7c0a753d66ebd194c0df108827f457442 and here 9b9342bddcd5dd78f0fbe021ba2bb16ef2695cd7

should be covered 😉

Grimothy:
thanks boss man

Issue #208, opened 2025-06-05
Notification of playlist channel or group changes.


craighton:
**Is your feature request related to a problem? Please describe.**
There is no easy way to track when channels or groups are added or removed from the source playlists.

**Describe the solution you'd like**
Would love some sort of tracking when channels or groups are added or deleted from playlists. 
I imagine either on the homepage a widget like box showing added or removed channels from the past sync or timeframe like 24-48 hours. 

My stretch request would be to have some sort of email notification or ntfy hook to broadcast when channels or groups have been added or removed from the playlist.

**Describe alternatives you've considered**

**Additional context**

![Image](https://github.com/user-attachments/assets/b2d819ea-3265-429e-adf3-9093cba841a5)

sparkison:
Already implemented! We don't send a notification, but do track each sync and the channels/groups removed/added.

Simply use the "Sync Logs" button on the Playlist edit page, or from the Playlist table view, use the three-dot menu and select the "View Sync Logs" action to be directed to the sync logs for the given Playlist 😉 

From Playlist edit/detail page:

![Image](https://github.com/user-attachments/assets/5739b07b-4b95-4bcb-9893-9fe906ce67f9)

From the table view:

![Image](https://github.com/user-attachments/assets/52cde06e-c1f4-4c35-8bfc-055f4703f9ba)

PR #1410, opened 2026-08-12
fix(epg): expose programme subtitle (episode name) on search_epg_shows, get_short_epg, get_epg_batch

### LINKED ISSUE #1409: Expose EPG programme subtitle (episode name) in search_epg_shows, get_short_epg, get_epg_batch
### Problem

EPG programme "episode name" (XMLTV `<sub-title>` — e.g. "Groundhog Day" for a *The Bear* airing, distinct from the show title "The Bear") is parsed and stored server-side, but none of the client-facing endpoints that serialize EPG programme data include it. Every consumer (TV app EPG grid, Shows search, etc.) can only distinguish airings of the same show by channel/time, never by episode name.

### Confirmed: the data exists, it's just never sent

- `EpgCacheService.php` correctly parses XMLTV `<sub-title>` into a `subtitle` field during ingestion (`case 'sub-title': $programme['subtitle'] = trim($reader->readString() ?: '');`, ~L225-226), including fallback extraction from description text via `EpgProgrammeNormalizer` when a provider doesn't tag it properly.
- It's persisted to the DB: `epg_programmes.subtitle` column exists (`database/migrations/2026_04_06_000005_create_epg_programmes_table.php`) and is written on ingest (`EpgCacheService.php` ~L1637).
- The JSONL programme cache also carries it end-to-end — the cache's own gap-filler placeholder arrays include a `'subtitle' => ''` slot (`EpgCacheService.php` ~L417, ~L791), confirming the cached format isn't missing the field either.

### The actual gap — three response builders drop it

`app/Http/Controllers/XtreamApiController.php`:

1. **`searchEpgShows`** (~L3851-3864) — queries the `EpgProgramme` Eloquent model directly (which has `subtitle`), but the `recentEpisodes` map only emits `channel_id/channel_name/title/start_time/end_time/season/episode/description`. `subtitle` is silently dropped.
2. **`get_short_epg`** action (~L1776-1789) — reads from the cache (which has `subtitle`), but the `$epgListings[]` build only emits `id/epg_id/title/lang/start/end/description/channel_id/start_timestamp/stop_timestamp/now_playing/has_archive`.
3. **`get_epg_batch`** action (~L2028-2040) — same gap; this is the endpoint behind the main Live TV EPG grid/timeline, so the gap isn't limited to DVR show search.

### Desired fix

Add the subtitle field to all three response arrays:
- `searchEpgShows`: `'subtitle' => $p->subtitle,` alongside the existing `title`/`description` keys in the `recentEpisodes` map.
- `get_short_epg` / `get_epg_batch`: `'subtitle' => $programme['subtitle'] ?? null,` (or base64-encoded, matching `get_epg_batch`'s existing `base64_encode($programme['title'] ?? '')` / `base64_encode($programme['desc'] ?? '')` treatment, for encoding parity with those two fields).

No schema/ingestion changes needed — this is purely "stop dropping a field that's already there."

### Downstream

TV client (`m3u-tv`) has a follow-up issue to consume and display this once available — will link once filed.

Grimothy:
Closes #1409.

## Bug

The XMLTV `<sub-title>` tag (the episode name, e.g. "Groundhog Day"
for a *The Bear* airing, distinct from the show title "The Bear") is
parsed into `epg_programmes.subtitle` during ingestion and cached, but
none of the three Xtream API endpoints that list programmes ever
included it in their responses:

- `searchEpgShows`'s `recentEpisodes` map
- `get_short_epg`'s `epg_listings` build
- `get_epg_batch`'s `epg_listings` build

## Fix

Add `subtitle` to all three response arrays, matching each endpoint's
existing encoding for `title`/`description`:

- `searchEpgShows`: plain string (`$p->subtitle`) — Eloquent-sourced,
  no encoding.
- `get_short_epg`: plain string (`$programme['subtitle'] ?? ''`) —
  cache-sourced, no encoding.
- `get_epg_batch`: base64-encoded (`base64_encode($programme['subtitle'] ?? '')`)
  — cache-sourced, matches this endpoint's existing base64 treatment
  of `title`/`description`.

These two encodings are intentionally different per endpoint (and
already were, for `title`/`description`) — not something to unify.

No schema, cache-format, or ingestion changes — the field already
exists everywhere upstream of these three response builders; this is
purely "stop dropping a field that's already there."

## Testing

New `tests/Feature/EpgSubtitleExposureTest.php` covers all three
endpoints, asserting `subtitle` round-trips correctly with the right
encoding for each. The `search_epg_shows` case is skipped under the
local `sqlite_testing` driver since that query path uses a
Postgres-only `ilike` operator (pre-existing, unrelated to this
change) — it runs under CI's Postgres driver.

`vendor/bin/pint --test` clean.

## Downstream

Unblocks the TV client follow-up: m3ue/m3u-tv#207.


Grimothy:
**Dependency graph for this EPG-subtitle work, across both repos:**

- **This PR (#1410)** — independent, no dependency on the other two. Safe to merge anytime.
- **m3ue/m3u-tv#208** — independent of this PR (different repo, different concern). Not blocked by this one.
- **m3ue/m3u-tv#210** — depends on **both**: it's git-stacked on #208 (must merge after it), and needs *this* PR merged for the episode-name data to actually be present (though it degrades safely to the show title if this hasn't shipped yet — no crash, just no new info).

No required merge order between this PR and #208 — either can go first. #210 should go last.

Grimothy:
**Update to the dependency graph:** a fourth PR joined this set — m3ue/m3u-editor#1412 (raises the `recent_episodes` cap and fixes its sort order so upcoming airings aren't buried, in the same `searchEpgShows` method this PR touches). Independent of this PR — no required order between the two, expect a routine rebase whichever lands second. Doesn't change anything else stated above.

PR #1412, opened 2026-08-13
fix(epg): prioritize upcoming episodes and raise the cap in search_epg_shows recent_episodes

### LINKED ISSUE #1411: search_epg_shows: recent_episodes sort order buries the next upcoming airing, cap too low for single-episode scheduling
### Problem

`searchEpgShows`'s `recent_episodes` field is capped at 5 airings per show and — more importantly — sorted in a way that doesn't serve the show-detail screen's actual use case anymore.

**Code** (`app/Http/Controllers/XtreamApiController.php`):

```php
// Sort within group: most recent first.
usort($progs, fn (EpgProgramme $a, EpgProgramme $b) => $b->start_time->timestamp <=> $a->start_time->timestamp);
...
$recentEpisodes = array_slice(array_map(...), 0, 5);
```

This sorts **descending** by `start_time` — i.e. the *furthest-in-the-future* (or most-recently-*past*, if there's no future data) airing comes first — matching the docstring's own description: `recent_episodes (up to 5 most-recent airings)`. `next_airing_at` is computed completely separately (its own loop finding the earliest *future* `start_time`, `next_airing_at: earliest start_time in the future, null if none`), so today `recent_episodes` and `next_airing_at` deliberately serve different purposes: `recent_episodes` for show identification/browsing, `next_airing_at` for "when's the next one" display.

### Why this needs fixing now

The TV client's Shows-detail screen (m3ue/m3u-tv#204) added a **per-episode "Record" button directly on each `recent_episodes` row** — so this list is no longer just for browsing/identification, it's the actual picker for which airing to schedule. Two compounding problems follow from the current design:

1. **Cap of 5 is low** for any show with several upcoming or recent airings (a daily/recurring show can easily exceed this within a typical EPG guide window).
2. **Sort order actively works against the new use case.** If a show has more than 5 future airings, the array is sorted furthest-future-first, so `array_slice(..., 0, 5)` can return the 5 *farthest-out* instances and silently exclude the very next upcoming episode — the one a user is most likely to want to record — from the list entirely.

### Proposed fix

- Reorder `$progs` so upcoming episodes come first, sorted ascending (soonest first), with past episodes appended after as a lower-priority tail, sorted descending (most recent past first) — same idea `next_airing_at` already uses for "what's next," just applied to the whole list instead of a single value.
- Raise the cap from 5 to a named constant (e.g. `20`) rather than a bare magic number.
- Update the docstring comment describing `recent_episodes` to match the new behavior/count.

### Explicitly out of scope

- No real pagination/cursor mechanism — the existing global 500-programme query cap and 100-show result cap remain; this only changes ordering and the per-show slice size within those existing bounds.
- No client (`m3u-tv`) changes required to consume more items — `ShowDetailScreen` already renders whatever `recentEpisodes` list length it receives.

Grimothy:
Closes #1411.

## Bug

`recent_episodes` was sorted **descending** by `start_time` (furthest-
future, or most-recent-past if no future data, first) and capped at 5.
That made sense when this list was purely for show identification/
browsing, but the TV client's Shows-detail screen (m3ue/m3u-tv#204,
PR m3ue/m3u-tv#208) now puts a per-episode **"Record"** button directly
on each row — making this list the picker for which airing to schedule.

A show with more than 5 future airings could have every slot filled by
the *farthest-out* instances, silently excluding the very next upcoming
episode — the one most likely to be recorded — from the list entirely.

## Fix

- Reorder into upcoming-first (soonest first), with past episodes as a
  lower-priority tail (most-recent-past first) — the same priority
  `next_airing_at` already uses for "what's next," now applied to the
  whole list instead of one value.
- Raise the cap from a bare `5` to a named `MAX_RECENT_EPISODES = 40`
  constant, matching the class's existing `private const` style.
- Update the docstring describing `recent_episodes` to match.
- Swap the query's Postgres-only `ilike` for a portable
  `whereRaw('LOWER(title) LIKE LOWER(?)', ...)`. Confirmed free — no
  index exists on `title` in `epg_programmes`, so both forms are
  equally unindexed leading-wildcard scans, zero performance
  difference. This closes a real gap: `search_epg_shows` previously had
  **zero test coverage exercised under the local `sqlite_testing`
  driver** — every prior test short-circuited at the 403 auth gate
  before reaching this query. CI runs Postgres and would have caught a
  regression, but nobody could verify locally before this change.

`next_airing_at`'s own computation is untouched — already correct and
independent of this change.

## Testing

New `tests/Feature/SearchEpgShowsOrderingTest.php`: 4 tests / 18
assertions covering upcoming-first ordering, past-as-tail, the raised
cap returning the *soonest* N (not the farthest — the actual bug,
verified with a defensive assertion that the old-sort behavior would
fail), and the no-upcoming-episodes fallback. Also ran the existing
`XtreamDvrOwnershipTest` (56 tests, 142 assertions) to confirm the
query change doesn't regress the access-control paths that also touch
this action. `vendor/bin/pint --test` clean.

## Related work / merge order

Independent of #1409 / #1410 (which adds `subtitle` to this same
response, unrelated concern) — no required order between this PR and
that one, they touch the same method but don't depend on each other,
expect a routine rebase whichever lands second.

Once this merges, m3ue/m3u-tv#210's stacked branch and #1410's
sqlite-skipped `search_epg_shows` test case can each simplify slightly
on their next rebase, but that's for those PRs to handle, not blocking
here.

### DIFF
--- a/app/Http/Controllers/XtreamApiController.php
+++ b/app/Http/Controllers/XtreamApiController.php
@@ -90,6 +90,15 @@ class XtreamApiController extends Controller
     ];
 
     /**
+     * Max number of `recent_episodes` returned per show in `search_epg_shows`.
+     * The list is the picker for which upcoming airing to schedule a recording
+     * on (m3u-tv#204), so it has to be large enough that the next actionable
+     * airing isn't pushed off the end by a deep schedule, while still bounding
+     * payload growth.
+     */
+    private const int MAX_RECENT_EPISODES = 40;
+
+    /**
      * Xtream API request handler.
      *
      * This endpoint serves as the primary interface for Xtream API interactions.
@@ -227,7 +236,8 @@ class XtreamApiController extends Controller
      * (bool — whether a Series rule already exists for this title), `series_rule_id` (int|null
      * — the existing rule's id when `has_series_rule` is true, for delete-without-round-trip),
      * `channel_count`, `channels` (array of `{channel_id, channel_name}`), `episode_count`,
-     * `next_airing_at` (ISO 8601 or null), `recent_episodes` (up to 5 most-recent airings).
+     * `next_airing_at` (ISO 8601 or null), `recent_episodes` (up to MAX_RECENT_EPISODES
+     * airings — upcoming first soonest, then most-recent-past).
      *
      *
      * @param  string  $uuid  The UUID of the playlist (required path parameter)
@@ -3797,7 +3807,7 @@ class XtreamApiController extends Controller
         $cutoff = Carbon::now()->subHours(24);
         $programmes = EpgProgramme::whereIn('epg_channel_id', $epgChannelIds->toArray())
             ->where('start_time', '>=', $cutoff)
-            ->where('title', 'ilike', '%'.$q.'%')
+            ->whereRaw('LOWER(title) LIKE LOWER(?)', ['%'.$q.'%'])
             ->limit(500)
             ->get();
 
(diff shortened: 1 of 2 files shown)


Grimothy:
**Dependency graph for the EPG-subtitle + episode-scheduling work across both repos:**

- **This PR (#1412)** — independent of m3ue/m3u-editor#1410 (touches the same method, unrelated concern — no required order between the two, expect a routine rebase whichever lands second).
- **m3ue/m3u-tv#208** — independent of this PR (different repo). Not blocked by this one, doesn't block this one.
- **m3ue/m3u-tv#210** — data dependency on this PR: its per-episode Record button is the reason this fix matters, but #210's code itself doesn't require this PR merged first (git-stacked on #208 instead). Works either merge order relative to #210.

No required merge order between this PR, #1410, or #208. All three can land in any order; #210 is the only one with a hard prerequisite (#208, git-wise).

Issue #1024, opened 2026-04-13
"Add to Custom Playlist" - Request timeout


ROPiator:
## Describe the bug

I use an Xtream playlist that includes a large amount of VOD content (_approx. 45k videos in 105 groups_). When trying to add a larger number of VOD groups to a "Custom Playlist" at once (_via multi-selecting categories_), I noticed a request timeout (_HTTP 504 Gateway Time-out_) that leads to issues with the added categories. Even though an error message ("Error while loading page - There was an error while attempting to load this page. Please try again later") is shown in the frontend, it seems like the categories are being added to the "Custom Playlist", but not with their (updated) "**Name**". In my case these categories are added with their "**Default name**"...

Additionally the "Add to Custom Playlist" dialog remains visible during the entire process, blocking the rest of the application and requiring it to wait until the process is finished. As a "bonus" it would be great if the process could happen in the background, notifying the user when the categories have been added (analog to the "Find & Replace" feature)...

## To Reproduce

1. Go to 'VOD Channels' > 'Groups',
2. Select a large number of categories by using the multi-select feature
3. Click 'Bulk actions' > 'Add to Custom Playlist'
4. Select the custom playlist and 'Use Original Item Group'
5. Click on 'Add now'
6. Wait for the error to be triggered when the request times out...

## Expected behavior

No error message is shown via the frontend when batch-adding a large number of VOD categories to a custom playlist and the categories are being added to the Custom Playlist with their correct display name.

## Actual behavior

The "Add to Custom Playlist" dialog remains visible and in the foreground (_blocking any further interaction with M3U editor_), an error message is shown after the timeout occurs, the dialog is not dismissed automatically and the categories are added with a wrong "Name" (their "Default name").

## Screenshots

<img width="427" height="213" alt="Image" src="https://github.com/user-attachments/assets/aead3a95-c708-492f-bc80-bbc5521a6484" />

"Blocking" dialog while the categories are being added

<img width="393" height="112" alt="Image" src="https://github.com/user-attachments/assets/b2a2333d-745e-41f7-b393-11885ae95e26" />

Error message shown via frontend

<img width="284" height="133" alt="Image" src="https://github.com/user-attachments/assets/8063f454-0e32-442e-a4e0-efd2c80b4e76" />

Error details for the respective HTTP request

<img width="1279" height="213" alt="Image" src="https://github.com/user-attachments/assets/37ab95a1-a475-4800-8bbf-ee8c8483ed19" />

[...]

PR #1427, opened 2026-08-17
fix(custom-playlist): move bulk add/detach to chunked background jobs

### LINKED ISSUE #1426: Detach VOD from custom playlist 504 timeout
## Describe the bug

When trying to detach a lot (100k+ in my case) of VOD channels from a custom playlist, the request times out with 504. I think this is similar to #1024.

## To Reproduce

 
1. Attempt to detach thousands of channels from a custom playlist
2. Request times out after 1.5 min

## Expected behavior

As with #1024 this should probably be a processing job.

## Actual behavior

Attempting to detach sits on the screen for ~1.5min before timing out with the same error as #1024.

Interestingly, I could not find anything in the compose, laravel or postgres logs. Only the error message in the UI and using dev tools to see the response.

ethan-manny:
## Summary

Fixes #1426 and addresses the two follow-ups from the last comment on #1024: bulk-adding **individually selected** channels still ran synchronously (504 on large selections), and the group-add background job was extremely slow / timed out on 100k+ channel libraries.

### What changed

**Detach (#1426)** — "Detach Selected" on the custom playlist Live/VOD/Series relation managers now dispatches a new `DetachItemsFromCustomPlaylist` job. It removes the playlist's group tags and pivot rows with chunked set-based deletes and notifies on completion, instead of looping per record inside the HTTP request.

**Individual adds (#1024 follow-up)** — "Add to Custom Playlist" on the Channels/VOD/Series resources dispatches a new `AddItemsToCustomPlaylist` job. The relation managers' "Add to custom group/category" actions had the identical timeout shape and use the same job.

**Group adds chunked (#1024 follow-up)** — `AddGroupsToCustomPlaylist` is now a thin parent that resolves each group's tag and item IDs, then fans out parallel `AddItemsToCustomPlaylistChunk` jobs via `Bus::batch` (same pattern as `ProcessChannelScrubber`). Tags are pre-created in the parent because `Tag::findOrCreate` is not race-safe across parallel workers.

**The real speedup** is in the new `PlaylistService::syncItemsToCustomPlaylist()`: the old per-item `detachTags()`/`attachTag()` loops (~4 queries × N items) are replaced with `insertOrIgnore`/bulk-delete queries (~6 queries per 1,000 items) that preserve existing pivot `channel_number`/`sort` values, following the `AutoSyncGroupsToCustomPlaylist` precedent. `addItemsToPlaylist()` now delegates to it, so there is a single implementation.

**Request-side memory** — the converted bulk actions use `fetchSelectedRecords(false)` and pluck selected IDs query-side, so "select all" over 100k+ rows no longer hydrates every model during the request.

~~**Queue config** — required because the jobs now fan work out via `Bus::batch()` instead of processing it inline: tests that call `handle()` and assert database state need the dispatched chunk jobs to actually execute. With the connection hardcoded to `redis`, phpunit.xml's `QUEUE_CONNECTION=sync` was ignored and the chunk jobs were pushed to a real Redis queue where nothing runs them. `config/queue.php` now honors `QUEUE_CONNECTION` **only when `APP_ENV=testing`**, so the sync driver executes the real batch pipeline inline. Production/dev keep the forced `redis` from 75edaf4c — deployments with a stale `QUEUE_CONNECTION` in `.env` are unaffected.~~ Reverted by 92920baae435fb2b5f2bdf4a1e176376bffcb92b

## Testing

- `AddGroupsToCustomPlaylistTest` — 14 passed (pre-existing suite, unmodified; now exercises the batch fan-out inline)
- `AddItemsToCustomPlaylistTest` (new) — 7 passed: select/create/original modes, per-item original group names, tag replacement, pivot value preservation, cross-playlist tag isolation, series categories
- `DetachItemsFromCustomPlaylistTest` (new) — 3 passed: pivot + tag removal, unselected/unrelated-tag isolation, series
- `AutoSyncGroupsToCustomPlaylistTest` — 19 passed (adjacent code, regression check)

## Security scan

`composer audit` and `npm audit` are clean. `php artisan checkpoint:scan` reports pre-existing repo-wide findings (driver-specific `orderByRaw` tag-sort clauses, DVR `proc_open`, media-server proxy TLS bypass, unguarded models by design via `Model::unguard()`); none are on lines added by this PR — all new queries use the query builder with bound parameters.

Both attaching and detaching channels of all types is now much quicker - it previously took 30min+ to maybe add all my VOD channels to a custom playlist, now it happens within seconds.


sparkison:
## Review Summary

This PR needs changes before merge. Found 4 correctness bugs, 3 duplication/reuse issues, 2 efficiency issues, and 1 style violation. Details below.

---

### Correctness

#### 1. Missing `->allowFailures()` on the Add batch
**File:** `app/Jobs/AddItemsToCustomPlaylist.php:91`

`Bus::batch()` is dispatched without `->allowFailures()`. If any single chunk job throws, Laravel cancels the entire batch, and every not-yet-run chunk silently no-ops (`if ($this->batch()?->cancelled()) return;`). For a 100k-item add split into ~100 chunks, one failure (deleted channel, transient DB error, lock timeout) can silently drop up to 95% of the items with only a generic failure notification shown to the user.

This PR's own cited precedent, `ProcessChannelScrubber`, calls `->allowFailures()` on its equivalent batch. `AddGroupsToCustomPlaylist` shares `dispatchChunkedBatch()` and inherits the same bug.

#### 2. `DetachItemsFromCustomPlaylist` has no retry/catch/failure notification
**File:** `app/Jobs/DetachItemsFromCustomPlaylist.php:43`

`tries = 1`, no try/catch, no failure `Notification`. A mid-run exception (deadlock, dropped connection) leaves the detach partially applied with zero user-visible feedback. The sibling `Add` job has `->catch()` → `notifyFailed()`; this one doesn't.

#### 3. Detach is still a single serial job, contradicting its own docblock
**File:** `app/Jobs/DetachItemsFromCustomPlaylist.php:53`

The docblock claims this design "cannot hit an HTTP or job timeout," but there's no `Bus::batch` fan-out here (unlike Add) — it's ~100 iterations of sequential DB round-trips in one worker with a 30-minute ceiling. A large enough playlist (or a slow/contended DB) can still time out, reproducing the exact bug class this PR set out to fix.

#### 4. `"0"`-named group/category is treated as falsy
**File:** `app/Services/PlaylistService.php:1025` (also line 1049, and `app/Jobs/AddGroupsToCustomPlaylist.php:59`)

`elseif ($tagName)` / `if ($tagName)` skips tag creation/application when `$tagName === "0"`, since PHP treats that string as falsy. A custom group literally named `"0"` gets synced into the playlist pivot but never tagged. Pre-existing bug, preserved through this rewrite.

---

### Duplication / Reuse

#### 5. `AddGroupsToCustomPlaylist` re-implements tag precreation instead of using the new shared helper
**File:** `app/Jobs/AddGroupsToCustomPlaylist.php:59`

This PR adds `PlaylistService::ensureCustomPlaylistTagsExist()` specifically to centralize race-safe tag precreation before parallel chunk jobs run. `AddGroupsToCustomPlaylist::handle()` hand-rolls the identical logic instead of calling it. Any future fix to that helper (e.g. the `"0"`-falsy bug above) won't propagate here.

#### 6. Regressed from `chunkById()` to unbounded `pluck('id')`
**File:** `app/Jobs/AddGroupsToCustomPlaylist.php:83`

Was memory-bounded, DB-side chunked reads via `chunkById(1000, ...)`; now `$group->$relation()->pluck('id')` loads every ID for the group into memory in one query. This is a step backward on exactly the scale this PR targets (100k+ channels per its own docblock), and violates the repo's memory-efficient-query standard (`->cursor()`/chunked reads over unbounded `->get()`/`->pluck()`). The sibling job `AutoSyncGroupsToCustomPlaylist` still does this correctly.

#### 7. `DetachItemsFromCustomPlaylist` duplicates the notification pair instead of reusing it
**File:** `app/Jobs/DetachItemsFromCustomPlaylist.php:63`

This PR extracts `notifyCompleted()`/`notifyFailed()` on `AddItemsToCustomPlaylist` for exactly this broadcast+`sendToDatabase` pattern. `DetachItemsFromCustomPlaylist` (added in the same PR) inlines its own copy instead — and, per 2, never wires up the failure path at all.

---

### Efficiency

#### 8. Group-name lookup query runs twice per item in "original" mode
**File:** `app/Services/PlaylistService.php:1038`

`ensureCustomPlaylistTagsExist()` runs `itemIdsByOriginalGroupName()` once (pre-dispatch, ~100 queries for a 100k-item/100-chunk add) just to learn tag names, discarding the id buckets it computes. Each of the ~100 parallel chunk jobs then re-runs the identical query inside `syncItemsToCustomPlaylist()`. Roughly doubles this step's query count for data already computed once.

#### 9. New jobs don't pin an explicit queue
**File:** `app/Jobs/AddItemsToCustomPlaylist.php:91`

None of the new jobs (`AddItemsToCustomPlaylist`, `AddGroupsToCustomPlaylist`, `DetachItemsFromCustomPlaylist`, `AddItemsToCustomPlaylistChunk`) call `->onQueue(...)`, so they land on Horizon's default queue. The cited precedent `ProcessChannelScrubber` pins to `'import'`. Inconsistent with the repo's queue-tagging convention and makes these jobs harder to monitor/prioritize independently in Horizon.

---

### Style

#### 10. Em dash in new docblock
**File:** `app/Services/PlaylistService.php:988`

`"* can run from parallel chunk jobs — with the caveat that any tag it needs"` uses an em dash, violating the workspace rule against em/en dashes in any string, comment, docblock, or commit message. Replace with a plain hyphen or rephrase.

---

### Recommendation

Closing this PR pending in favor of a different solution.

PR #1439, opened 2026-08-20
Fix series NFO provider IDs and artwork fields

### LINKED ISSUE #1438: Series NFO omits dedicated provider IDs and artwork fields
## Bug description

`NfoService::generateSeriesNfo()` reads provider IDs and artwork only from legacy keys in the `metadata` JSON payload. Current `Series` records also store these values in dedicated columns such as `tmdb_id`, `tvdb_id`, `imdb_id`, `cover`, and `backdrop_path`.

When metadata enrichment populates the dedicated columns but not the legacy JSON keys, the generated `tvshow.nfo` omits valid provider IDs and poster or backdrop URLs. Media servers can then fail to identify the series or display a primary poster even though m3u-editor already has the data.

## Steps to reproduce

1. Create or load a `Series` with dedicated provider columns and `cover` populated.
2. Leave the corresponding legacy `metadata.tmdb_id`, `metadata.tvdb_id`, `metadata.imdb_id`, and `metadata.poster_path` keys absent.
3. Generate `tvshow.nfo` through `NfoService::generateSeriesNfo()`.
4. Observe that the provider IDs and `<thumb>` are missing.

## Expected behavior

Series NFO generation should use the current model fields first and retain legacy metadata JSON as a fallback:

- provider IDs should follow `Series::getMovieDbIds()`
- poster artwork should use the dedicated `cover` field before legacy image keys
- backdrop artwork should use the dedicated `backdrop_path` field before legacy image keys
- existing legacy metadata payloads must remain compatible

## Actual behavior

The generator reads only legacy metadata keys, so valid current model data is silently omitted.

## Acceptance criteria

- Dedicated provider columns are emitted into `tvshow.nfo`.
- Episode NFOs use the same current series provider fields instead of stale JSON-only lookups.
- Dedicated poster and backdrop fields are emitted into `tvshow.nfo`.
- Legacy metadata-only records continue to produce the same IDs and artwork.
- Regression tests cover dedicated-field precedence and legacy fallback behavior.

Serph91P:
## Summary

- use dedicated series provider ID columns before legacy metadata
- use dedicated cover and backdrop fields before legacy artwork metadata
- apply the same provider fallback to episode NFO generation
- preserve compatibility with older metadata-only records

## Verification

- `./vendor/bin/pint --test app/Services/NfoService.php tests/Unit/NfoServiceTest.php`
- `REVERB_APP_ID=test REVERB_APP_KEY=test REVERB_APP_SECRET=test php artisan test tests/Unit/NfoServiceTest.php`
- 17 tests passed with 40 assertions
- regression sabotage confirmed the new provider tests fail against the previous implementation
- `git diff --check`

Closes #1438

### DIFF
--- a/app/Services/NfoService.php
+++ b/app/Services/NfoService.php
@@ -29,9 +29,9 @@ class NfoService
         try {
             $metadata = $series->metadata ?? [];
 
-            $tmdbId = $this->getScalarValue($metadata['tmdb_id'] ?? $metadata['tmdb'] ?? null);
-            $tvdbId = $this->getScalarValue($metadata['tvdb_id'] ?? $metadata['tvdb'] ?? null);
-            $imdbId = $this->getScalarValue($metadata['imdb_id'] ?? $metadata['imdb'] ?? null);
+            $tmdbId = $this->getSeriesProviderId($series, 'tmdb');
+            $tvdbId = $this->getSeriesProviderId($series, 'tvdb');
+            $imdbId = $this->getSeriesProviderId($series, 'imdb');
 
             $xml = $this->startXml('tvshow');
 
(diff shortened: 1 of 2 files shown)


Serph91P:
Hi @sparkison, this fix is finished, verified, and ready for review and merge.

PR #1448, opened 2026-08-22
Fix Provider Pooling (Profile) stream ID mapping when pool connects via m3u-editor's own Xtream API

banshee2005:
## Problem
Provider Pooling (Provider Profiles) fails when the pool playlist connects through m3u-editor's own Xtream API instead of directly to the IPTV provider. This affects providers that only support M3U (no Xtream API), e.g. EPGenius.

## Root Cause
The pool playlist syncs from m3u-editor's own Xtream API, which returns stream_id = the SOURCE playlist's channel PK. When transformUrl() swaps to a secondary profile whose password is the TARGET playlist's UUID, m3u-editor authenticates as the TARGET playlist and looks up the stream ID there - but that PK only exists in the SOURCE playlist, so it returns 403 / "Failed to load stream".

## Fix
- PlaylistProfile::transformUrl() now calls resolveProviderStreamId() to map the stream ID to the channel PK in the playlist the profile authenticates against, via the shared source_id (MD5 set during M3U import).
- New PlaylistProfile::resolveProviderStreamId() maps: source PK -> source channel -> source_id -> target channel -> target PK.
- XtreamStreamController::getValidatedStreamFromPlaylist() now logs which playlist authenticated and whether the channel was found, to aid diagnosis.

## Behavior
- Direct-to-provider setup: unchanged (returns null).
- Primary profile: unchanged.
- Secondary profile via indirection: stream ID remapped to target playlist's channel PK.
- Channel genuinely absent in target playlist: logs warning, passes through (403).

## Verification
- Primary URL -> HTTP 302 -> stream (works)
- Secondary, unmapped ID -> HTTP 403 (the bug)
- Secondary, mapped ID -> HTTP 302 -> valid MPEG-TS streamed (works)

### DIFF
--- a/app/Http/Controllers/XtreamStreamController.php
+++ b/app/Http/Controllers/XtreamStreamController.php
@@ -17,6 +17,7 @@ use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Redirect;
 
 class XtreamStreamController extends Controller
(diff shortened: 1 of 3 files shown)


sparkison:
I see what ya did there! Clever! Haha! It wasn't meant to be used this way, so that's why you're running into issues. At the same time, we don't guard against inception 😜 ...I'm going to review, and possibly take a different angle. But I'll make sure to get a fix in regardless...

Issue #1382, opened 2026-08-09
Channel Merging Duplicates


ethan-manny:
## Describe the bug

When merging is enabled for a playlist that contains two or more channels with the same id but the same stream URL, one of these channels (I think the first one the merge comes across) will acquire failovers to the other channels - even though the failovers point to the same stream.

## To Reproduce

1. Enable merging for a playlist that contains two or more channels with the same id but same stream URL.
2. Sync & process that playlist.
3. Upon completion check the duplicate channels and observe that the one with failover(s) has a failover(s) that point to the same stream.

## Expected behavior

When merging, if the channel candidate has the same URL then it should not become a failover.

## Actual behavior

When merging, if the channel candidate has the same URL then it will not become a failover.

Serph91P:
@sparkison I think this is a bit of a tricky one and would "bloat" up the merge. What do you think should we implement a check if the stream url are the same or make it optional atleast?

sparkison:
> [@sparkison](https://github.com/sparkison) I think this is a bit of a tricky one and would "bloat" up the merge. What do you think should we implement a check if the stream url are the same or make it optional atleast?

I think we should pass the master channel through and run a duplicate check on the normalized URL before assigning as failovers. It makes sense and I don't feel it would add a ton of overhead.

I'll see what I can figure out...

sparkison:
Got it! This will be fixed in the next release (https://github.com/m3ue/m3u-editor/commit/b3eae2979c1369ed8960a803cba0e5efdc36ef6a)

sparkison:
fixed in latest release: [v0.12.49](https://github.com/m3ue/m3u-editor/releases/tag/v0.12.49) 🎉

PR #1477, opened 2026-09-03
feat: add optional title similarity guard to channel merging

gabelul:
**Draft on purpose.** There's a question at the bottom I'd rather have answered before you spend time reviewing an implementation. The bug is real either way; the fix is a proposal.

## The bug

`MergeChannels` groups channels by `stream_id` and writes a `ChannelFailover` for every other member of the group. Apart from the URL-duplicate check from #1382, nothing checks what those channels are — the merge key is taken entirely on faith.

That's fine until the key isn't unique, and on an Xtream playlist `stream_id` holds the provider's tvg-id — `ProcessM3uImportChunk.php:122` documents it as *"provider stream ID or tvg-id ... for Xtream API this could be `epg_channel_id`"*.

On my install the provider returns the literal string `TS` in `epg_channel_id` for 82 channels:

```
name           = UK: BBC 3 / CBBC HD ◉
stream_id      = 162100
epg_channel_id = TS          <-- provider's value, stored as channels.stream_id
```

The editor stores that faithfully and keeps the real ID in `source_id`. All 82 therefore share one merge key and grouped as a single channel. `UK: BBC 3 / CBBC HD` came out with **81 failovers**, among them Food Network +1, HGTV +1, ITV London 4K, Daystar, Film4 and Horror Xtra. When BBC 3 drops, the viewer gets Food Network.

VOD is worse, because nothing about it is defensible: *Pitch Perfect 3* failing over to *Tenéis que venir a verla*, *Brimstone* to *Doctor Strange*.

Providers cause the same shape of problem on their own — nine Sky Sports **Darts** channels here ship with `tvg-id="SkySportsF1.uk"`, so Darts became a backup for F1.

Across 12,754 failovers on my install, 534 paired channels that have nothing to do with each other.

**Correction:** an earlier version of this description blamed the importer for parsing the URL. That was wrong — I inferred it from these URLs ending in `.ts` and never checked the payload or the import path. There is no URL parsing; the value comes straight from the provider. Details and the raw payload are in the comments below.

The wider pattern is that `stream_id` is not unique on any of my playlists while `source_id` is:

| Playlist | Channels | distinct `source_id` | distinct `stream_id` |
|---|---|---|---|
| A | 45,416 | 45,416 | 37,158 |
| B | 12,371 | 12,371 | 1,233 |
| C | 31,212 | 31,212 | 26,056 |

Every collision group holds fully distinct `source_id`s. Grouping on the tvg-id is deliberate — it's the cross-provider identity, and `source_id` is provider-local — but it means the merge key is a field providers fill with whatever they like.

## What this proposes

A `minTitleSimilarity` option on `MergeChannels`. When set, a candidate has to agree with its master on title as well as on merge key.

Being straight about what it is: **a heuristic filter, not a compatibility guarantee.** It can flag a pairing as implausible. It cannot establish that two channels *are* the same. It defaults to `0.0` (off), so nothing changes unless someone opts in, and it abstains whenever it can't judge confidently.

Comparing raw titles doesn't work — provider titles are mostly not the channel name, and a source prefix plus quality decorations dominate any comparison. `DE| SYFY HEVC` against `DE| SYFY FHD` is the same channel and scores 0.47. So titles reduce to a bare core name first, and only the cores get compared.

## Three worked examples at threshold 0.40

Real pairs from a live install, not constructed:

**Caught** (487 of 534):
```
UK| TRUE CRIME HD    vs  UK: CBS REALITY            cores truecrime / cbsreality     sim 0.32
DE| RTL HEVC         vs  DE| BILD+ EVENT FAME ...   cores rtl / bildeventfame...     sim 0.08
```

**Missed** (47 — kept, shouldn't have been):
```
IT| REALTIME UHD     vs  IT: SKY RETE 7             cores realtime / skyrete7        sim 0.50
IT| SKY DAZN 1 UHD   vs  IT - DAZN CHANNEL HEVC     cores skydazn1 / daznchannel     sim 0.42
```

**Wrongly rejected** (21 — legitimate pairs refused):
```
DE: LEAGUES FOOTBALL PPV 1 - NO EVENT...  vs  ENDED | FV RAVENSBURG VS SSV REUTLINGEN...  sim 0.32
```
That one is a PPV channel against the fixture it's currently carrying. The titles share almost nothing and no title comparison recovers it.

## Full numbers

| Threshold | Correctly kept | False positives | Bad pairs caught | Missed |
|---|---|---|---|---|
| 0.30 | 12,205 | 15 (0.12%) | 448 | 86 |
| **0.40** | **12,199** | **21 (0.17%)** | **487** | **47** |
| 0.50 | 12,113 | 107 (0.88%) | 498 | 36 |
| 0.55 | 11,951 | 269 (2.20%) | 514 | 20 |

0.40 is the knee. Verified across all 12,754 pairs that `0.0` changes nothing.

**The number that actually matters for a decision** isn't the 0.17% false-positive rate against everything kept — it's the collateral rate among what gets blocked. At 0.40 the guard blocks 508 pairs, and **21 of those (4.1%) are legitimate**. So roughly one in twenty-four blocks is wrong. That's the tradeoff, stated the honest way.

## What it can't do

Same-brand siblings defeat it, and I don't think any scoring approach fixes them. Sky Sports Darts against Sky Sports F1 scores 0.53. `BBC One (Scotland)` against `BBC One (London)` about 0.6. `BBC NEWS: WORLD` against `CNN NEWS: WORLD` about 0.83, since they share nine of twelve characters. Real overlapping text, different channels.

Two names both under five characters are abstained on rather than guessed at.

Nothing here repairs a bad `stream_id`, and existing bad failovers are untouched — this only affects merges run afterwards.

## Review history

The first version had nine defects, found by an adversarial review and all reproduced before being fixed. Two were the kind that make a feature actively harmful: the guard was skipped entirely when the master had no URL, and bracketed text was stripped so `The Thing (1982)` and `The Thing (2011)` both normalised to `thething` — the guard *accepted* one film as the other's failover. Each has a regression test now, and the hardened version beats the first on both precision and recall.

One approach tried and dropped: scaling the score by the ratio of the two name lengths. Principled-looking, fixed two edge cases, and cost a tenfold worse false-positive rate on the same dataset, because legitimate pairs like `NAT GEO WILD` and `NAT GEOGRAPHIC WILD` differ in length by design.

## Running it

```bash
vendor/bin/pest tests/Unit/MergeChannelsTest.php tests/Unit/ChannelTitleSimilarityServiceTest.php
vendor/bin/pint --test app/Services/ChannelTitleSimilarityService.php app/Jobs/MergeChannels.php
```
61 tests, 96 assertions, Pint clean locally.

**CI note:** Security Scan fails on four `league/commonmark` advisories published 2026-09-01. This branch doesn't touch `composer.json` or `composer.lock` — both are byte-identical to master and the branch is 0 commits behind. Tests and Docker Build are gated behind Security Scan, so **the test job has never run on this PR**. Everything above is from local runs.

## The question

Do you actually want this behaviour upstream?

I built it because it bit me, but it's unsolicited, it's domain-specific, and a heuristic with known blind spots is a maintenance burden someone else ends up carrying. Three ways this could go, and I'm happy with any of them:

1. You want it roughly as-is, and I take review feedback.
2. You'd rather have a simpler compatibility invariant than a similarity score — tell me the rule and I'll implement that instead.
3. It's too specific for upstream and this closes. The `'TS'` importer parse is still worth a look on its own, and I'll open that as a plain issue with a fixture.

### DIFF
--- a/app/Jobs/MergeChannels.php
+++ b/app/Jobs/MergeChannels.php
@@ -8,6 +8,7 @@ use App\Models\Group;
 use App\Models\Playlist;
 use App\Models\User;
 use App\Services\ChannelMergeKeyService;
+use App\Services\ChannelTitleSimilarityService;
 use Filament\Notifications\Notification;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
@@ -82,6 +83,7 @@ class MergeChannels implements ShouldQueue
         public string $contentType = 'live',
         public string $mergeKey = 'stream_id',
         public bool $scrubberAwareMasterSelection = false,
+        public float $minTitleSimilarity = 0.0,
     ) {
         $this->contentType = in_array($this->contentType, ['live', 'vod'], true) ? $this->contentType : 'live';
         $this->mergeKey = in_array($this->mergeKey, ['stream_id', 'tmdb_id'], true) ? $this->mergeKey : 'stream_id';
@@ -89,6 +91,10 @@ class MergeChannels implements ShouldQueue
         if ($this->contentType !== 'vod' && $this->mergeKey === 'tmdb_id') {
             $this->mergeKey = 'stream_id';
         }
+
+        // Clamp to a sane range. 0.0 keeps the guard off, which is the default
+        // so existing merges behave exactly as before.
+        $this->minTitleSimilarity = max(0.0, min(1.0, $this->minTitleSimilarity));
     }
 
     /**
@@ -550,6 +556,11 @@ class MergeChannels implements ShouldQueue
      * master's - a failover pointing at the exact same stream the master
      * already plays provides no redundancy, it's just a no-op duplicate.
      */
+    /**
+     * Drop failover candidates that would be useless or actively wrong:
+     * anything pointing at the master's own URL, and — when enabled — anything
+     * whose title says it is a different channel.
+     */
     protected function filterFailoverCandidates(Collection $group, Channel $master): Collection
     {
         $masterUrl = $this->resolveEffectiveUrl($master);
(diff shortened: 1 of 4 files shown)


sparkison:
Closing! This is feeling like more of an exploration as apposed to a well-scoped PR. Since it's in flux and changing I'm going to close in favor of opening a more targeted and confirmed working PR 😉

Issue #1391, opened 2026-08-09
Playlist bouquets


ethan-manny:
**Is your feature request related to a problem? Please describe.**
When using playlist aliases, it would be good if we could select a pre-defined set of groups to include. Currently for each alias, you have to select the groups manually.

**Describe the solution you'd like**
Create "bouquets" which are a user-defined selection of the playlist's groups. For example a user could create a USA bouquet that contains all groups with USA channels. Then, when creating a playlist alias there is an option to select which bouquet to assign - as well as the existing group selection.

**Describe alternatives you've considered**
The same functionality can be achieved using custom playlists. However, I'd argue this feature is more streamlined if implemented.

PR #1506, opened 2026-09-11
fix: Merged-target bouquet pickers reopen empty and drop selections on save

ethan-manny:
## Summary

- Reopening a bouquet that targets a **merged playlist** showed empty group/category pickers, so every edit meant re-selecting everything. Saving that form then wiped the stored selections.
- Standard-playlist and custom-playlist targets are not affected.
- Regression introduced with the merged-playlist target added on #1484 (`d1fb4f8a`); the original #1391 branch had no merged targets, which is why it didn't show there.

## Root cause

Filament applies `OptionsArrayStateCast::set()` to a multi `ModalTableSelect`'s state *before* `afterStateHydrated` runs, and that cast fails closed on non-scalar items. The merged picker stores `{playlist_id, name}` pairs, so the hook always saw `$state === []` and bailed out. With no ids in state, `dehydrateStateUsing` then read the save as a deliberate clear (its never-silently-shrink guard only keeps entries that no longer resolve) and persisted `[]`.

## Fix

`MergedSourceGroupModalSelect::afterStateHydrated` now reads the pairs from `$record->group_selections[<key>]`, the same pattern `PlaylistAliasResource::persistedSourceSelection()` already uses for merged aliases, and skips when the state is already populated (a re-fill). `pairsToSourceIds()` drops its re-normalisation, which is redundant now that its only caller passes normalised pairs.

## Tests

`tests/Feature/BouquetEditHydrationTest.php`: reopen for standard, custom and merged targets (asserting picker state and the rendered picker labels), plus save-after-reopen for a merged target. The two merged cases fail without the fix. Ran together with `BouquetResourceTest` and `BouquetMergedPlaylistTest`: 19 passed. Pint clean.

## Checklist notes

- No user-facing strings added or changed, so no `lang/` updates.
- `checkpoint:scan`: the same pre-existing buckets as `dev` (hardcoded secrets in lang files, `orderByRaw`, `GitInfoService` `shell_exec`, `MediaServerProxyController` TLS); nothing in the changed files. No dependency changes.
- Merged-target bouquets that were saved from the broken edit form have already lost their selections and need re-picking once; nothing here can recover them.
- Not changed here (pre-existing, unrelated to merged targets): `SourceGroupModalSelect` decides whether to translate names with `is_string($state[0])`, so a standard-target bouquet whose first stored group name is all digits would not be translated to ids.

### DIFF
--- a/app/Filament/Forms/Components/MergedSourceGroupModalSelect.php
+++ b/app/Filament/Forms/Components/MergedSourceGroupModalSelect.php
@@ -83,20 +83,26 @@ class MergedSourceGroupModalSelect
                     ? SourceCategory::displayLabelsForIds($playlistIds, $ids, includePlaylistName: count($playlistIds) > 1)
                     : SourceGroup::displayLabelsForIds($playlistIds, $type, $ids, includePlaylistName: count($playlistIds) > 1);
             })
-            ->afterStateHydrated(function ($component, $state, $record, Get $get) use ($scopedQuery, $playlistIdsFor): void {
+            ->afterStateHydrated(function ($component, $state, $record, Get $get) use ($scopedQuery, $selectionKey, $playlistIdsFor): void {
                 // Hidden twin components are still hydrated; bail unless this is a
-                // merged-target record whose stored pairs need resolving to ids.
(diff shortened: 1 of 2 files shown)


sparkison (commented):
Looks good! ❤️

PR #1487, opened 2026-09-05
feat(sort): add 'Sort by Rating' option to playlist Sort Alpha Configs

Grimothy:
## Summary

Adds the ability to sort **VOD channels** and **Series** by TMDB rating from the playlist settings → **Sort Alpha Configs** Repeater. Mirrors the existing "Sort by Release Date" option end-to-end.

A new `Rating` entry appears in the **Sort By** dropdown when the target is `vod_groups` or `series_categories`. Selecting it defaults the sort direction to `DESC` (highest rated first), matching the natural user expectation. Unrated content sinks to the bottom regardless of ASC/DESC.

### `app/Services/SortService.php` — 4 new methods

| Method | Scope |
|---|---|
| `bulkSortGroupChannelsByRating` | VOD within a single group |
| `bulkSortCategorySeriesByRating` | Series within a single category |
| `bulkSortPlaylistVodByRating` | All VOD in playlist, globally (1..N, no collisions) |
| `bulkSortPlaylistSeriesByRating` | All Series in playlist, globally (1..N, no collisions) |

All use `ROW_NUMBER()` window functions on MySQL / Postgres / SQLite with a PHP-side `CASE`-update fallback for other drivers — same pattern as the existing `*ByReleaseDate` methods.

### Rating sources

- **VOD channels:** `info->rating` JSON (TMDB `vote_average`). The dedicated `channels.rating` column was [dropped in the 2025_06_23 migration](database/migrations/2025_06_23_205558_add_container_extension_column_to_channels.php) — the JSON value is now the only source. The expression casts to `DECIMAL(4,2)` for ordering.
- **Series:** direct `rating` column (declared as `string` in [2025_05_21](database/migrations/2025_05_21_185309_create_series_table.php) but stores decimal values like `"7.5"`). Cast to `DECIMAL` for ordering.

### NULL ordering

Unrated rows always sink to the bottom, regardless of `$order`. Critical implementation detail: the `IS NULL` check uses the **raw extraction** (no `COALESCE`/`NULLIF` wrapping), so unrated rows are still detected even when the value expression is wrapped in `COALESCE → 0` for ordering. Without this, an ASC sort would float unrated items to the top — never what the user wants for "sort by rating".

### `app/Jobs/RunPlaylistSortAlpha.php`

Dispatches the new methods when `sort_alpha_config.column === 'rating'`, mirroring the existing `release_date` branches:

- `target = vod_groups` + `column = rating` + `all`: global playlist VOD sort
[...]

PR #1510, opened 2026-09-12
feat(vod-merge): resolution-aware prioritization for same-TMDB movies

Grimothy:
## Summary

Adds the ability to prioritize higher-resolution duplicates when the same TMDB ID appears at multiple resolutions in a VOD playlist. The higher-resolution copy becomes the master and lower-resolution copies become ChannelFailover children.

Driven by `auto_merge_config.merge_key === 'tmdb_id'`; gated by the new `vod_resolution_priority_enabled` toggle (default: on when opted in).

### Resolution source

- **VOD channels**: `info->rating` JSON (TMDB `vote_average`). The dedicated `channels.rating` column was [dropped in the 2025_06_23 migration](database/migrations/2025_06_23_205558_add_container_extension_column_to_channels.php). The expression casts to `DECIMAL(4,2)` for ordering.
- **New**: `VodResolutionExtractor` reads probed `stream_stats` first, falls back to parsing resolution from title/name/URL via `AioStreamsQualityParser`. Never live-ffprobes — purely read-only over stored data.

### Scoring

- New vertical-resolution normalization in `MergeChannels`: 4K (2160p) = 100, 1080p = 50, 720p = 33, 480p = 22. Replaces legacy pixel-count scoring (1080p = 25, 720p = 11) only when `vod_resolution_priority_enabled` is on.
- Legacy pixel-count scoring preserved for `vod_resolution_priority_enabled = false` paths.

### New `auto_merge_config.vod_*` toggles (gated by `merge_key=tmdb_id`)

| Key | Default | Effect |
|---|---|---|
| `vod_resolution_priority_enabled` | `true` | Use vertical-resolution scoring for VOD merges |
| `vod_use_filename_resolution` | `true` | Parse resolution from title/name/URL when ffprobe data missing |
| `vod_min_resolution_promote` | `720` | Discard filename-derived resolution below this height |
| `vod_verify_filename_via_probe` | `false` (opt-in) | Queue unresolved channels for ffprobe after merge completes |

### Pipeline

- Hooks into existing `MergeChannels` engine (`contentType='vod'`, `mergeKey='tmdb_id'`). No new job class — `VodResolutionExtractor` is the only new file in `app/Services/`.
- `vod_verify_filename_via_probe=true` dispatches `ProbeStreamsChunk` for the unresolved channels post-merge; next merge uses the probed value.

## Behavior change (for existing users)

Existing users with `auto_merge_config.merge_key=tmdb_id` and no explicit `vod_resolution_priority_enabled` value will get the new resolution-priority scoring turned on automatically. Vertical-resolution normalization differs from legacy pixel-count:
- 1080p: 25 → 50
- 720p: 11 → 33

Ranking changes only affect which master/failover pair gets chosen when multiple same-TMDB-ID copies are present at different resolutions. The intended UX is "highest resolution wins" — this is the documented intent.

## Bug fixes addressed in PR (in addition to the feature)

- **H1** (silent bug): `vod_verify_filename_via_probe` was being silently dropped by `SyncListener::buildWeightedConfig` and `PlaylistController::buildMergeWeightedConfig` — both now emit all four `vod_*` keys, matching `PlaylistService::buildMergeWeightedConfig` (the manual-modal path). The auto-merge path now actually dispatches the verify probe.
- **M2** (UX): `auto_merge_config.merge_key` Select in `PlaylistResource.php` and the manual-modal `merge_key` Select in `PlaylistService::getMergeFormSchema` now have `->live()`, so the VOD Resolution Priority fieldset's `hidden()` closure re-evaluates when the user flips `merge_key`.
- **L1** (i18n): All `->label()` / `->helperText()` strings in the VOD Resolution Priority modal fieldset are now wrapped in `__()`; new keys added to `lang/en.json`.
- **N1** (UX consistency): The VOD Resolution Priority fieldset in the manual modal now has `hidden(fn (Get $get) => $get('merge_key') !== 'tmdb_id')`, matching the sibling fieldset in PlaylistResource.

## Tests

New tests:
- `tests/Unit/VodResolutionExtractorTest.php` — 4 cases
- `tests/Unit/MergeChannelsVodResolutionTest.php` — 6 cases
- 2 new assertions in `tests/Feature/SyncListenerTest.php` covering `vod_verify_filename_via_probe` opt-in (true) and default (false)

```
vendor/bin/pest tests/Unit/VodResolutionExtractorTest.php tests/Unit/MergeChannelsVodResolutionTest.php tests/Feature/SyncListenerTest.php
Tests: 47 passed (78 assertions)
```

## Pre-PR checklist

- [x] **Security scan** — `composer audit --locked` → 4 pre-existing `league/commonmark` advisories (no new findings introduced; same as PR #1487).
- [x] **Lang files** — 9 new keys added to `lang/en.json` via `app:sync-translations`; `php artisan lang:merge-conflicts` → no conflicts.
- [x] **Pint** — `vendor/bin/pint` on changed files → passed.
- [x] **Tests** — scoped test files above all pass.

## Out of scope (not addressed in this PR)

- M3 (consolidate 3 `buildMergeWeightedConfig` implementations into one — pre-existing duplication; tracked for a future refactor PR).


sparkison:
## Review Summary

Found two confirmed critical issues plus a few plausible follow-ups. Details below.

### Critical

**1. `vod_verify_filename_via_probe` is a no-op unless an unrelated toggle is also enabled**
`app/Jobs/MergeChannels.php:229` (`dispatchVodResolutionProbeVerification`)

```php
if (! $playlist || ! $playlist->auto_probe_vod_streams) {
    return;
}
```

Enabling `auto_merge_config.vod_verify_filename_via_probe` in the new VOD Resolution Priority fieldset does nothing unless the playlist's separate `auto_probe_vod_streams` setting ("Probe VOD & series streams after sync", default off, lives in a different section) is also on. No log or notification indicates why the probe never fires. This contradicts the toggle's own label and the new docs, which describe it as a standalone opt-in.

**2. `vod_resolution_priority_enabled` defaults to `true`, silently changing existing merge behavior**
`app/Listeners/SyncListener.php:253`

For any existing playlist with `merge_key = tmdb_id` and a custom `priority_attributes` order, the next sync will default `vod_resolution_priority_enabled` to `true` and force-prepend `resolution` onto that order (via `PlaylistService::vodPriorityAttributes()`), changing which VOD duplicate becomes master with no explicit opt-in. This looks like it should default to `false` to match the "no behavioral regression without opt-in" convention used elsewhere in this repo's merge/sync job chains.

**3. Em dash used in code and docs**
`app/Jobs/MergeChannels.php:226` and several sentences in `docs/auto-merge-channels.md`

```php
Log::warning("MergeChannels: Failed to queue VOD resolution verification probe - {$e->getMessage()}");
```

Repo convention disallows em/en dashes anywhere, including log strings and docs. Please swap for a hyphen or rephrase.

### Worth a look

- **`VodResolutionExtractor::extract()`** (`app/Services/VodResolutionExtractor.php:31`) reads raw `title`/`name`/`url` but ignores `title_custom`/`name_custom`/`url_custom`, which every other comparison in `MergeChannels.php` prefers when present. Could cause resolution to be extracted from the wrong string.
- **`hydrateVodResolutions()`** (`app/Jobs/MergeChannels.php:704`) only caches resolutions for the post-filter `$eligibleGroup`, but failover scoring later pulls candidates from the original unfiltered group. A channel excluded by `exclude_disabled_groups`/`scrubber_aware_master_selection` but still eligible as a failover would have no cache entry and silently score 0 on resolution.
- **Locale files**: new strings were added only to `lang/en.json`. Per the project's pre-PR checklist, `de.json`/`es.json`/`fr.json`/`zh_CN.json` should be updated too (`php artisan lang:merge-conflicts` after).
- **Duplicated ternary**: `($config['merge_key'] ?? 'stream_id') === 'tmdb_id' ? 'vod' : 'live'` appears verbatim at `PlaylistController.php:530` and `:545`, and again in `SyncListener.php:237`/`:241`. Worth extracting to a shared helper to avoid drift.
- **Docs**: `docs/auto-merge-channels.md` references `screenshots/vod-resolution-priority.png`, which isn't included in the PR.

sparkison:
@Grimothy 

still seeing some issues...

1. PlaylistResource.php:2698 — the auto-merge vod_resolution_priority_enabled toggle **defaults to true in the UI**, but the manual modal, the runtime ?? false fallback, and the PR's own test all say **it should default false**. Editing an existing playlist can silently flip this on and change merge behavior.
2. MergeChannels.php:468 — the regex-merge pass's query drops stream_stats from the select list, **so probed-resolution data is invisible to that pass and scoring silently falls back to filename parsing**.
3. PlaylistController.php:416 — the manual merge API's validation rules and config-override loop never picked up the four new vod_* keys, **so per-request overrides are silently dropped** — the same bug class (H1) the PR claims to have fixed, just in a different call path.
4. PlaylistService.php:513 — the vod_* config-assembly block is **duplicated verbatim across three call sites instead of using a shared helper**, despite the PR already introducing helpers for exactly this pattern. This is likely why finding 3 happened.
5. VodResolutionExtractor.php:33 — container resolution and full group re-hydration **repeat on every merge pass**, adding avoidable overhead to a hot job-chain path.

PR #1512, opened 2026-09-13
feat(series,vod): link actor filmography from series and VOD detail views

Grimothy:
## Summary

When a Series or VOD detail view renders the rich cast row (populated by upstream's TMDB enrichment via `$record->metadata['cast_list']`), each actor tile is now wrapped in a link to **ActorFilmography** filtered to the originating playlist. The filmography page then shows only items available locally — i.e. the same scope the existing ArrSearch cast section already uses.

This is the in-app side of the same data shape the future m3u-tv app integration will consume (the 'door' the user asked to leave open).

## What's in the diff (rebased onto upstream `experimental` @ 2b326c34)

- **Cast display partial:** `cast-section.blade.php` is extended (from upstream) with optional `filmographyPage` / `playlistId` / `playlistUuid` params. When supplied, each tile becomes a link to the actor's filmography. When not supplied, the original static upstream behavior is preserved.
- **`view-series.blade.php` + `view-vod.blade.php`:** detect the current panel and pass the right `filmographyPage` class + `playlistId` (admin) or `playlistUuid` (guest) to `cast-section.blade.php`.
- **Trait:** `FiltersFilmographyByPlaylist` shared between admin + guest filmography pages — holds `filterFilmographyToPlaylist()` + `resolveLocalItemUrl()`. `cursor()` + hash-set lookup gives O(1) per credit and flat memory regardless of playlist size.
- **`ActorFilmography` + `GuestActorFilmography`:** optional `playlistId` / `playlistUuid` query param. When set, TMDB combined credits filter to Series/Vod that exist locally; clicking a filmography item redirects to the local Series/Vod detail (admin vs guest resource class).
- **Admin ownership check:** admin's `playlist_id` param is validated against the current user (or admin). Foreign playlist_ids silently fall back to global TMDB.

## Builds on upstream

This PR extends **`2b326c34 feat: Extend TMDB enrichment to include related`** (already on `experimental`). The cast display, TMDB enrichment pipeline, clearLogo, and rating-suppression work are all upstream's; this PR adds only the navigation layer + playlist-scoped filmography page behavior.

The implementation deliberately reads cast from upstream's `metadata['cast_list']` (cached on the model) instead of fetching live on every view load — that pattern would have added a TMDB HTTP call per page render, capped the visible cast at 15 (instead of upstream's ~50 via `aggregate_credits`), and burned through the TMDB rate limit.

## Tests added (99 passing in affected scope, 176 assertions)

- `tests/Feature/CastSectionPartialTest.php` (new, 4 cases) — empty, rendered, with link wrapper, static fallback. Covers the extended contract of upstream's partial.
- `tests/Feature/ActorFilmographyTest.php` — playlist scope, no-filter on zero, foreign-playlist rejected (non-admin), redirect to local Series.
- `tests/Feature/GuestActorFilmographyTest.php` — existing populate test re-seeded to seed a matching VOD for the new filter.

## Decisions / notes

- **Scope** is playlist-scoped filmography only. Global TMDB filmography is preserved as the default when the param is absent — same call site the existing ArrSearch guest flow uses.
- **Auth.** Guest side inherits the playlist from `HasGuestAuth` (no extra check needed).
- **No migrations.** §5 of pr-review-standards is N/A.
- **No `Process*Import` / `Sync*` job chain touched** — change gets extra scrutiny per the project's PR review standards.

## Checklist

- [x] Pint clean on all modified files
- [x] Tests green (99 / 99 in the affected scope; full test suite should re-verify on CI)
- [x] `php artisan lang:merge-conflicts` — no conflicts
- [x] No migrations
- [x] Rebased onto current `experimental` (this PR is now based on 2b326c34)

## Heads-up: pre-existing security scan finding

`composer audit` flags 4 high-severity advisories in `league/commonmark <2.9.1`. Current install is 2.9.0; advisories fixed in 2.9.1. **Pre-existing on `experimental`**, not introduced by this branch.


sparkison:
Sick! Looks great, nice job. Reviewing now...

PR #1517, opened 2026-09-14
fix: route genuine HLS sources for series and VOD through the HLS proxy

jWXZSCsqrpZRSVwauBjn:
With proxying enabled, genuine HLS sources for series/VOD are currently forced through `/stream/{id}`. Master playlists reach the client unchanged, causing it to request private backend addresses and fail with “Source error”.

Replace the HLS-to-raw overrides with content-aware proxy routing for ambiguous on-demand URLs (`.m3u8`, `.hls` and extensionless sources). Apply this consistently to new and reused streams, including the channel capacity-reuse path.

The proxy inspects its normal playback response, so the editor makes no additional provider request. Genuine HLS uses playlist rewriting; video files with misleading `.m3u8` extensions remain playable. Known video formats, DASH, live routing and transcoding retain their existing paths.

Reproduction: configure a series/VOD source whose HLS master references relative variants and absolute audio/subtitle playlists on an internal-only backend. Enable proxying and play it.

Validation:
- 12 regression cases passed, 36 assertions: series/VOD, new/reused streams and HLS-shaped/video/opaque URLs.
- Pint passed.
- `composer security:scan` passed: zero Composer/npm vulnerabilities; 26 Checkpoint checks passed without warnings.

Requires the companion m3u-proxy MR: https://github.com/m3ue/m3u-proxy/pull/61
Deploy the proxy change first. Older proxies do not support the binary fallback on this endpoint.


sparkison:
After digging in more, I'm finding this PR builds an entirely new branch for the proxy when it could be done much more simply. Going to close in favor of a more targeted fix...

Issue #1518, opened 2026-09-15
Schedules Direct sync completes without EPG programme data


Serph91P:
## Bug description
The Schedules Direct import can reportedly complete without importing any EPG programme data. This report is based on an affected user instance; no reproducible runtime instance or Schedules Direct response/log bundle is available to the reporter yet, so the root cause is not asserted here.

## Current code path checked
On `dev` at `2d09743eb4e69bd7a7ea006ed11e5204fc779f6c`, `App\Services\SchedulesDirectService::syncEpgData()` authenticates, loads the configured lineup, requests `/schedules`, collects `programID` values, requests `/programs`, writes XMLTV `<programme>` elements, and then `ProcessEpgImport` parses the generated XMLTV file. The service catches/logs per-sub-chunk program-fetch exceptions and continues, which can leave a generated XMLTV file with channels but no programmes while the overall sync path proceeds. This is a code observation, not a confirmed root cause.

## Expected behavior
A successful Schedules Direct sync should contain programme entries in the generated XMLTV and the subsequent import should report a non-zero programme count when the provider returned schedule/program data. If the provider returns no data or an API error, the sync should fail visibly with the relevant response/status rather than appear successful with an empty programme set.

## Actual behavior
The affected instance reportedly has no EPG programme data after import. Reproduction and the exact failing stage are currently unknown.

## Needed runtime evidence
Please attach, with credentials/tokens redacted:
- m3u-editor version and commit;
- import start time and EPG configuration (days, lineup/station count);
- logs covering `/token`, `/lineups/{lineup}`, `/schedules`, and `/programs`, including HTTP status/provider `code` and response sizes;
- generated XMLTV channel/programme counts (or a redacted sample);
- resulting import/cache counts and any `sd_errors`/job error.

## Scope
This issue is for the canonical upstream repository and does not claim that an existing commit fixes a running user instance.

sparkison:
PR reviewed and commented on 😉

Issue #1420, opened 2026-08-15
Support multiple SchedulesDirect lineups per EPG


Serph91P:
## Problem

A SchedulesDirect EPG can currently be configured with only one lineup. However, the lineup management dialog reports the account usage as, for example, `1 of 4 slots used`, which implies that the remaining account slots can also be used from the EPG configuration.

## Current behavior

- The EPG form stores and selects a single SchedulesDirect lineup.
- The management dialog can list and remove account lineups and displays the account's actual lineup limit.
- There is no way to add further lineups to the same EPG even when the SchedulesDirect account has free slots.

## Expected behavior

Allow a SchedulesDirect EPG to use multiple lineups up to the account limit reported by the API. The UI should make the selected lineups and remaining slots clear and consistent across creation, editing, and lineup management.

## Acceptance criteria

- Multiple SchedulesDirect lineups can be selected for one EPG, up to the account limit returned by SchedulesDirect.
- Import processing includes stations and schedule data from every selected lineup.
- Users can add and remove individual lineups without unintentionally affecting the others.
- The displayed slot usage reflects the actual selected/account lineups and updates after changes.
- Existing single-lineup EPG configurations continue to work without manual migration.
- Attempts to exceed the account limit or API lineup-change limits show a clear error.

## Related context

The account lineup count and removal workflow were introduced around #1196, but EPG configuration still remains limited to one lineup.

PR #1519, opened 2026-09-15
Fix SchedulesDirect imports and manage lineups

Serph91P:
Fixes #1518. Implements #1420.

- Store ordered multi-lineup EPG selections with legacy single-lineup fallback and primary synchronization.
- Validate account membership and account limits before persisting EPG selections, while allowing account and EPG selections to be managed separately.
- Import every selected lineup through the documented encoded endpoint, deduplicating overlapping stations while retaining atomic XMLTV replacement on a failed batch.
- Preserve partial-batch diagnostics, documented transient retries, and existing SchedulesDirect request streaming.

Tests are HTTP-mock based; no live SchedulesDirect credentials or mutations used.


sparkison:
@Serph91P 

1. addLineup() targets a nonexistent endpoint (app/Services/SchedulesDirectService.php:607) — changed to PUT /lineups with {lineup: id} in the body, but SchedulesDirect's real API is PUT /lineups/{lineupID} with an empty body. This breaks both the new "Add Lineup" admin action and the pre-existing auto-add-on-sync fallback — likely a bigger regression than the bug it's trying to fix.
2. Partial batch failures now corrupt the XMLTV output file (line ~1485) — the loop now throws on any failed program batch, but that happens after the old file was already deleted and the new one opened, so the exception skips the closing </tv> write. Consumers get a truncated/invalid file instead of the previously-complete one.
3. A deleted catch(Exception) (line ~1630) means any transient error (network blip, bad date, parse error) — not just the intended SD 6000/6001 codes — now escalates to a full sync failure instead of skipping one batch. Combined with 2, one hiccup can now corrupt the whole EPG.
4. New em dash in EpgResource.php:907 ("{$lineup['name']} — {$lineup['lineup']} ...") violates the repo's explicit no-em-dash rule — minor, easy fix.

None of the test files added actually exercise the real HTTP calls, so 1 in particular would sail through CI while failing in production.

Serph91P:
> @Serph91P
> 
>     1. addLineup() targets a nonexistent endpoint (app/Services/SchedulesDirectService.php:607) — changed to PUT /lineups with {lineup: id} in the body, but SchedulesDirect's real API is PUT /lineups/{lineupID} with an empty body. This breaks both the new "Add Lineup" admin action and the pre-existing auto-add-on-sync fallback — likely a bigger regression than the bug it's trying to fix.
> 
>     2. Partial batch failures now corrupt the XMLTV output file (line ~1485) — the loop now throws on any failed program batch, but that happens after the old file was already deleted and the new one opened, so the exception skips the closing  write. Consumers get a truncated/invalid file instead of the previously-complete one.
> 
>     3. A deleted catch(Exception) (line ~1630) means any transient error (network blip, bad date, parse error) — not just the intended SD 6000/6001 codes — now escalates to a full sync failure instead of skipping one batch. Combined with 2, one hiccup can now corrupt the whole EPG.
> 
>     4. New em dash in EpgResource.php:907 ("{$lineup['name']} — {$lineup['lineup']} ...") violates the repo's explicit no-em-dash rule — minor, easy fix.
> 
> 
> None of the test files added actually exercise the real HTTP calls, so 1 in particular would sail through CI while failing in production.

Sorry about this one gonna fix this. Thank you

Serph91P:
@sparkison can you please take another look?

sparkison:
@Serph91P 

Still seeing some issues...

1. Add-lineup form is unsubmittable — [EpgResource.php:926](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Filament/Resources/Epgs/EpgResource.php#L926): only lineup_to_remove is ->required(). A user trying to just add a lineup gets blocked by client-side validation, and on a fresh account with zero existing lineups, lineup_to_remove has no options at all — the headline "add a lineup" feature can't be used standalone.
2. Dead error-recovery code — [SchedulesDirectService.php:1495](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Services/SchedulesDirectService.php#L1495): $programBatchFailures is declared but never appended to, so the "preserve valid data on partial batch failure" behavior described in the PR never actually triggers — any batch failure still aborts the whole sync.
3. Removed per-batch try/catch causes bigger data loss than before — [SchedulesDirectService.php:1637](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Services/SchedulesDirectService.php#L1637): one failing HTTP batch now skips every remaining batch in that sub-chunk instead of just logging and continuing, a regression from the prior log-and-continue behavior.
4. Inconsistent URL encoding between the two new lineup actions — [SchedulesDirectService.php:636](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Services/SchedulesDirectService.php#L636): removeLineup() still interpolates the raw lineup ID into the DELETE URL while the PR's own addLineup() was fixed to rawurlencode() it. The PR's own test fixture (USA-NY12345-X/plus) would break removeLineup.
5. Streaming optimization defeated on the happy path — [SchedulesDirectService.php:1654](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Services/SchedulesDirectService.php#L1654): fetchProgramBatch() now does file_get_contents on every response (success or not) just to check for an error code key, undoing the memory benefit of the ->sink()-based streaming this PR introduces.
6. Missing lang/ updates — [EpgResource.php:916](vscode-webview://0gipnolvea8koa346fg1bu5t0c8rk17dl17crgmhthr7q60vru56/m3u-editor/app/Filament/Resources/Epgs/EpgResource.php#L916): new user-facing strings ("Apply Lineup Change", "Lineup to Add", etc.) weren't added to lang/ for any locale, violating the repo's mandatory pre-PR checklist.

Serph91P:
@sparkison sorry going to leave these here I just can't concentrate right now. But there are issues with schedules direct.

Serph91P:
@sparkison Thanks for the review. All points from your comment are addressed in the current head.

- Add-only form: `lineup_to_remove` is required only when no add selection is made, so a pure add submits even with an empty remove list.
- Program batches: failures are collected per 5000-program batch, the remaining batches of the same subchunk are still requested and processed, and the partial failure is surfaced only after the full pass.
- Lineup DELETE: the lineup id is URL-encoded in the request path.
- Success path: the program response is streamed into the sink; the body is read only on the error path.
- Translations: the new strings are present in lang/en.json, alphabetically ordered, without duplicate keys.
- Regression tests: partial-failure coverage (first batch fails permanently, a later batch in the same subchunk is still processed, the error is surfaced, an existing XMLTV file stays unchanged) and URL-encoding coverage for a lineup id containing "/" are both present.

Test evidence on this head: the 6 affected test files run with 28 tests, 80 assertions, 0 failures.

CI on this head: all 4 checks are green (Code Quality, Security Scan, Tests, Docker Build Validation).

The PR is out of draft and ready for review.

Serph91P:
@sparkison Thanks again for the detailed feedback. The current head `b15735be8d1ff9f84b77073fca004d644be46624` addresses the review points and now implements the per-EPG multi-lineup behavior from #1420.

- The normal EPG picker exposes only lineups already present in the SchedulesDirect account. Headend discovery remains in the explicit account-management add flow, with slot information and read-back before selection.
- Existing single-lineup EPGs remain compatible; multiple selected lineups are persisted locally and synchronized through the documented one-lineup provider endpoints.
- Exhausted `/schedules` retries and failed program batches now surface diagnostics before XMLTV promotion. The previous XMLTV file is retained byte-for-byte on these failures.
- EPG deletion no longer mutates the SchedulesDirect account. Account-lineup removal is rejected while any same-account EPG still selects that lineup.
- Add/remove lineup paths URL-encode lineup IDs, and successful program responses remain streamed.
- Regression coverage includes add-only validation, URL encoding, multi-lineup selection and sync, legacy fallback, account-vs-EPG isolation, and atomic XMLTV preservation for exhausted schedule retries.

The current-head GitHub checks are all green: Code Quality, Security Scan, Tests, and Docker Build Validation.

Could you please take another look when convenient? Thank you.

PR #1520, opened 2026-09-15
fix: proxy logos in channel table

jWXZSCsqrpZRSVwauBjn:
## Summary

- load `enable_logo_proxy` with constrained playlist relations in the channel table
- ensure channel images use the configured logo proxy for regular and custom playlists
- add regression coverage for both playlist types

## Tests

- `php artisan test --compact tests/Feature/ChannelLogoProxyTest.php`


sparkison:
Nice find, looks good! 🙏  I just added coverage for VOD too (it had the same issue).

PR #1521, opened 2026-09-16
feat: add host-owned EPG enrichment API

Serph91P:
## Summary

- collapse the EPG cache to a single canonical file set per EPG: no generation directories, no pointer file, no retained copies
- apply enrichment patches in place: one write connection, `BEGIN IMMEDIATE`, revision checks and the revision bump inside the same transaction, and the commit is the publish
- keep the bounded snapshot read-only against the same file (pinned revision, `LIMIT`) instead of copying it
- invalidate the affected XMLTV and gzip outputs only after a successful publish
- keep v1 cache layouts readable but reject mutation on them

## What changed since the review

The earlier revision published immutable cache generations into a `generations/` directory with a pointer file. That duplicated `programmes.sqlite` on every cache run and on every successful patch - the storage growth you hit while testing. That design is gone:

- `EpgCacheStorage` owns one canonical file set per EPG (`metadata.json`, `channels.json`, `programmes.sqlite`, `CACHE_VERSION = v2`). A full rebuild is staged aside and published with an atomic `rename`; staged files are dropped on failure and nothing is retained per run.
- Applying enrichment patches is an in-place transaction against that same file: one writer and one `SELECT` per batch, `ERRMODE_EXCEPTION`, a busy timeout, `BEGIN IMMEDIATE`, row and revision checks before the updates, and the revision bump in the same commit. Readers see one side of the commit, never a mix.
- The bounded snapshot stays a read-only query with a pinned revision and a `LIMIT` - no file copy.
- `EpgCacheGenerationResolver` and its test are deleted, and the now-dead `EpgCacheService::clearCache()` was removed together with its orphaned private helper.

Your three findings:

| Finding | Now |
|---|---|
| A failed `UPDATE` silently no-ops while `apply()` still reports `applied` | `PDOException` is rethrown, and `apply()` only reports `applied` when the store reports it |
| Nothing deletes a superseded generation, so disk grows with every run | Generation copies and the pointer file no longer exist; two consecutive rebuilds leave the same file set at the same sizes (covered by a test) |
| Opening a connection per changed row let lock contention surface as `invalid_patch` | One writer and one `SELECT` per batch, and SQLite busy codes / lock timeouts map to a transient error |

## Follow-up fixes

Two follow-ups landed on top of the rework, both converging on the same rule - contention is retryable, it is never a verdict on the cache or on the plugin's batch:

- busy/lock is now classified as a transient error in the revision pre-check, in the row pre-check and in the snapshot (`EpgCacheBusyException` becomes `transient_error`; a locked read is no longer reported as `invalid_patch` or `legacy_cache_read_only`, and no longer escapes as an uncaught exception)
- `publishRebuild()` validates `metadata.json` and `channels.json` as a JSON object/array again before the first `rename`, so a truncated or scalar payload is not published into the canonical cache

## Storage behaviour

At rest there is exactly one cache file set per EPG, in the same layout `dev` uses today (`epg-cache/{uuid}/v2/`). While a rebuild is running, the new set is staged next to the old one before the atomic swap, so there is a transient peak of up to 2x for the duration of that rebuild; nothing accumulates across runs. Enrichment patches carry no such peak because they update the canonical file in place.

## Validation

- focused EPG cache and enrichment service tests: 41 passed, 171 assertions (`tests/Feature/EpgCacheStorageTest.php`, `EpgCacheInPlaceUpdateTest.php`, `EpgCacheEnrichmentServiceTest.php`, `EpgCacheProgrammeIndexTest.php`, `EpgCacheEnrichmentMutationLockTest.php`)
- `composer security:scan`: 26 passed, 0 failed
- CI on the final head `70fd0c66026641d6b7bf16fd757b0d8646f83880`: all four checks green (Code Quality, Security Scan, Tests, Docker Build Validation) - https://github.com/m3ue/m3u-editor/actions/runs/35354420499
- `php -l` on the changed PHP files and `git diff --check` clean
- diff against `dev`: 11 files, +2757/-136

## Scope

Host half of the EPG enricher migration. The companion plugin draft lives in `Serph91P/m3u-editor-epg-enricher`. No deployment included.

Closes #1526


Serph91P:
@sparkison could you please take a look at this too?

sparkison (commented):
Found a couple issues, I'll get addressed directly while I'm here:

1. Silent data loss: writeChanges() and readRows() open raw PDO connections without ERRMODE_EXCEPTION (unlike EpgProgrammeStore, which sets it explicitly). A failed UPDATE silently no-ops while apply() still returns 'applied' — plugins get told a patch succeeded when it wasn't written.
2. Unbounded disk growth: nothing deletes a superseded generation on success. EpgCacheService::clearCache() is now dead code (zero callers, confirmed via grep), and every scheduled EPG refresh plus every successful enrichment patch leaves a full programmes.sqlite copy on disk forever.
3. Lock contention mislabeled as bad input: hasExpectedRevisions()/prevalidatedChanges() open a new SQLite connection per changed row (up to ~200 for a 100-patch call), some of it inside the shared per-EPG 30s mutation lock. Contention can trip the 10s lock timeout, which gets reported back to the plugin as 'invalid_patch'.

sparkison:
@Serph91P gonna need to see if you can change how things are stored, re: the `generations` folder with a pointer file, etc. There needs to be a single source of truth here - one cache file, not copies/duplicates/etc.

I rebuilt the entire EPG cache to use less space and be more optimized, now this PR will double the storage (at a minimum! It's actually worse as it builds with each cache run) by duplicating the cache file. That's a non-starter for me unfortunately - we need to find another solution here.

I've added some fixes to other items I found, but got blocked after testing and seeing the cache files keep growing...

Serph91P:
> @Serph91P gonna need to see if you can change how things are stored, re: the `generations` folder with a pointer file, etc. There needs to be a single source of truth here - one cache file, not copies/duplicates/etc.
> 
> I rebuilt the entire EPG cache to use less space and be more optimized, now this PR will double the storage (at a minimum! It's actually worse as it builds with each cache run) by duplicating the cache file. That's a non-starter for me unfortunately - we need to find another solution here.
> 
> I've added some fixes to other items I found, but got blocked after testing and seeing the cache files keep growing...

@sparkison going to look at it. Thanks for the feedback!

Serph91P:
Reworked the storage side after your review: there is now a single cache file per EPG, with no generations directory and no pointer file.

- `EpgCacheStorage` owns the canonical file set. A full rebuild is built aside and published with an atomic `rename`; staged files are removed on failure, nothing is retained per run. Two consecutive rebuilds leave the same file set at the same sizes - covered by a test.
- Enrichment apply is in-place in that same file: one write connection with `ERRMODE_EXCEPTION` plus a busy timeout, `BEGIN IMMEDIATE`, revision checks inside the same transaction, revision bump in the same commit, and the commit is the publish. Readers see one side of the commit, never a mix.
- The bounded snapshot is still a read-only query against the same file with a pinned revision and a `LIMIT` - no file copy.

Your three findings:
- A failed UPDATE can no longer silently no-op: `PDOException` is rethrown and `apply()` only reports `applied` when the store actually reports it.
- The unbounded growth is gone with the design: generation copies and the pointer file are removed (`EpgCacheGenerationResolver` and its test are deleted, the now-dead `clearCache()` was dropped).
- Lock contention is no longer reported as `invalid_patch`: one writer and one `SELECT` per batch, and SQLite busy codes / lock timeouts now map to a transient error.

Head is `49c300aba`. CI on that commit is green - Code Quality, Security Scan, Tests, Docker Build Validation: https://github.com/m3ue/m3u-editor/actions/runs/35341232469. Focussed EPG cache/enrichment suite: 34 passed (156 assertions).

Marking this ready for review again.

Serph91P:
@sparkison The host-side EPG enrichment API is ready for another review. Your feedback has been incorporated: at rest, each EPG now has one canonical cache file set, enrichment updates are applied in place transactionally, and lock contention is handled as retryable/transient rather than invalid input. The final CI checks are green.

One clarification on the storage requirement: current `dev` clears the existing cache before rebuilding, so it never has a temporary duplicate. This PR keeps the previous canonical cache readable while it writes a staged replacement and atomically renames it into place. That prevents accumulation across runs, but creates a temporary peak of up to 2x during a rebuild. Did you mean one canonical cache at rest, or no temporary duplicate at all? If the latter is required, I will adjust this to the current `dev` clear-first behavior.

Thanks again for the detailed review.

PR #1524, opened 2026-09-18
feat(cache): add cached content downloads

Grimothy:
## What This Feature Is

Cached Content Downloads lets a user save provider-backed VOD movies and series episodes to local application storage. Once a file finishes downloading, playback automatically uses the local copy instead of repeatedly fetching the provider URL. If caching is disabled, the file is incomplete, expired, unavailable, or unauthorized, playback falls back to the existing live-stream path.

## How To Use It

1. Enable cached content in the integration settings.
2. Use **Cache Now** on a VOD movie, series episode, or supported dynamic-group member, or let the cache command discover eligible content.
3. Monitor progress from the cached downloads page/widget.
4. Play the content normally. Completed cache entries are selected automatically; users do not need a different playback URL.

Cache entries can be retained according to the configured policy. Playlist-level sharing controls whether the same content may be reused across playlists. Dynamic-group membership is tracked through pivots so a file can be retained while it remains relevant to at least one active group.

### 1. Eligibility And Dispatch

- Cache actions resolve the channel or episode's provider URL and canonical content identity.
- Channel and episode fingerprints are centralized so manual caching, scheduled caching, playback lookup, and retention use the same identity rules.
- Episode identity includes the parent series identity to avoid collisions between similarly numbered episodes.
- Existing completed/downloading rows are reused rather than creating duplicate downloads.
- Dispatch is guarded by the cache-enable setting and by the content type's cacheability.

Primary code: `CachedContentDispatchService`, `Channel`, `Episode`, `CachedContentFile`, the cache actions, and `CacheContentCommand`.

### 2. Queued Download

`DownloadCachedContentFile` claims the row atomically, downloads the provider response as a stream, writes it to the configured storage disk, and records progress and terminal state.

- Downloads are chunked instead of buffered in memory.
- Progress is persisted at throttled byte boundaries.
- Cancellation is cooperative between chunks; it does not throw from a curl progress callback.
- Failed downloads retain an error message and retry metadata.
- A completed write is validated before the row becomes `Completed`.
- Active downloads are protected from retention cleanup.

### 3. Storage And Security

Outbound provider URLs are validated by `PrivateNetworkGuard`, including redirect targets. The downloader rejects unsafe/private destinations before following them.

Cached files are written through the configured storage disk. Stream-write failures and database failures are allowed to surface instead of falsely marking a file complete.

### 4. Retention And Cleanup

Retention evaluates current content membership lazily rather than loading every candidate into memory. It avoids deleting active downloads and respects never-expire entries, playlist ownership, shared-cache settings, and dynamic-group pivot state.

Scheduled maintenance currently runs as:

- `cache:cleanup` at 03:00
- `cache:cleanup-orphans` at 03:30
- `cache:content --dry-run` hourly

`CachedContentRetentionCleanup` exists as a queued cleanup job; the scheduler currently invokes the cleanup command directly. These are related but distinct execution paths.

Primary code: `CachedContentRetentionService`, `CachedContentRetentionCleanup`, `CacheContentCleanupCommand`, and `CacheContentOrphanCleanupCommand`.

### 5. Playback

`XtreamStreamController` checks for an eligible completed cache entry before using the live provider redirect. It must not serve cached content when caching is disabled.

`CachedContentStreamController` performs playlist-credential ownership authorization before serving a file. The route is publicly reachable at the HTTP middleware level, so authorization inside the controller is mandatory. The controller delegates local-file delivery to `StreamLocalFile`.

`StreamLocalFile` supports full responses and valid byte ranges. Invalid ranges return `416 Range Not Satisfiable`; valid ranges return the appropriate partial response and headers.

## Security Invariants

- Never fetch or follow a provider URL that resolves to a blocked/private destination.
- Validate every redirect target, not only the initial URL.
- Never allow one user or playlist to retrieve another user's cached file.
- Do not serve a completed cache entry while caching is disabled.
- Do not expose an incomplete or failed file as playable content.
- Do not mark a file complete until the streamed write is valid.
- Preserve range-response correctness for media clients.
- Keep active downloads out of retention deletion.

## Maintainer Review Order

Review the feature as one lifecycle before splitting into parallel checks. Agents should first agree on these invariants, then inspect their assigned area:

1. **Data model and identity**: `CachedContentFile`, migrations, fingerprints, ownership, playlist sharing, and dynamic-group pivots.
2. **Dispatch**: manual actions, scheduled discovery, deduplication, cache-enable/type guards, and queue selection.
3. **Download worker**: atomic claiming, redirects, SSRF checks, streaming, progress, cancellation, retries, failures, and final-write validation.
4. **Retention**: expiration rules, never-expire behavior, active-download protection, lazy queries, shared playlists, dynamic groups, and orphan cleanup.
5. **Playback**: cache-hit selection, disabled-cache bypass, credential authorization, range handling, and live fallback.
6. **UI and operations**: settings, Cache Now actions, downloads page/widget, progress/error display, commands, scheduler, and Horizon queue configuration.
7. **Tests and regressions**: map each invariant to an existing focused test; flag uncovered paths rather than assuming adjacent tests cover them.

## Validation Checklist

- Confirm a cache request creates one row and one queued download for a cache miss.
- Confirm repeated requests reuse the existing row and do not duplicate work.
- Confirm channel and episode fingerprints are stable and distinct where required.
- Confirm disabled caching prevents both new dispatch and cached playback.
- Confirm a completed cache hit redirects to the authorized local stream.
- Confirm missing, failed, expired, or unauthorized cache entries fall back to live playback.
- Confirm unauthorized playlist credentials cannot access another user's cache.
- Confirm unsafe initial URLs and unsafe redirect targets are rejected.
- Confirm cancellation stops between chunks and leaves a non-playable terminal state.
- Confirm stream-write failure cannot produce a `Completed` row.
- Confirm progress and error fields are persisted at the expected boundaries.
- Confirm valid ranges work and invalid ranges return `416`.
- Confirm retention does not remove active or never-expire files.
- Confirm dynamic-group pivot rows are attached, soft-unshared, and considered by retention.
- Confirm orphan cleanup removes storage/database orphans without deleting valid files.
- Confirm scheduled commands and queue assignments match the documented operational paths.

## Current Verification

- Cache-focused run: **202 tests passed, 1 failed (480 assertions reported)**.
- The remaining failure is `tests/Feature/DynamicGroupResourceTest.php`: the test expects no `edit` action on the Dynamic Groups Movies relation manager, but Filament currently exposes an `EditAction`.
- `vendor/bin/pint --dirty --format agent` passed.
- PHP syntax checks and `git diff --check` passed.
- Composer security audit passed. `npm audit` remains blocked by HTTP 400 because the local package tree/lockfile is invalid.

## Known Coverage Gaps To Check

- Dedicated tests for the orphan-cleanup command.
- Direct coverage for per-playlist retention and sharing UI overrides.
- Explicit redirect-walk coverage across safe and unsafe targets.
- Explicit `PrivateNetworkGuard` coverage for all supported URL forms.
- Resolution of the Dynamic Groups relation-manager `edit` action mismatch before merge.


sparkison:
@Grimothy 

## Review notes for #1524 (cache download feature)

Automated multi-angle review across correctness, security, reuse, efficiency, and conventions. Grouped by severity, file:line references are against `pr-1524`.

### Correctness - high priority

1. **Cross-playlist cache sharing is dispatched but never served** (`app/Services/CachedContentDispatchService.php:232`, `app/Http/Controllers/XtreamStreamController.php:571`)
   `dispatchNew()` skips creating a new cache row whenever *any* playlist already has that fingerprint, with no `playlist_id` or `share_cache_across_playlists` check. But `resolveCacheHit()` only serves `Completed` rows scoped to the requesting playlist's own `playlist_id`. Net effect: playlist B silently never gets its own cached copy when playlist A already cached the same fingerprint, and even when `share_cache_across_playlists` is intentionally enabled, the shared file is never actually served at playback time.

2. **Cancelled downloads get stuck at "Downloading" forever** (`app/Jobs/DownloadCachedContentFile.php:179`, `:239`)
   Both cancellation checks return from `handle()` without setting `status` to `Failed`. Horizon reports the job as successful and never retries, and `atomicReclaim()` only reclaims `Pending`/`Failed` rows, so the row is a permanent zombie requiring a manual DB edit.

3. **Fingerprint collision for unmatched content** (`app/Models/CachedContentFile.php:128`, `Channel.php:217`, `Episode.php:115`)
   When `tmdb_id`/`tvdb_id` are both null, the fingerprint collapses to a constant string (e.g. `"movie::::"`). Two different unmatched movies on the same playlist can collide, causing the wrong cached file to be served for the second one.

4. **New `cache_retention_mode` playlist setting is never read** (`app/Services/CachedContentRetentionService.php`)
   The PR adds this column and wires it into the Playlist form, but the retention service never imports `Playlist` or checks the field, only `never_expire` and live-fingerprint membership. Setting a playlist to `manual`/`never-expire` retention mode is a silent no-op, files still get swept.

5. **`enable_cache` kill switch not enforced everywhere** (`app/Console/Commands/CacheContentCommand.php`, `CachedContentDispatchService.php`, `DownloadCachedContentFile.php`)
   Only checked in UI action visibility and the two serve controllers. Running `php artisan cache:content --playlist=42` (the documented ad-hoc invocation) while "Enable Cache" is off in Settings still dispatches real downloads.

6. **"Cache Now" shows a false failure notification** (`app/Filament/Resources/Vods/VodResource.php:2540`, `EpisodesRelationManager.php:492`)
   `dispatchForChannel`/`dispatchForEpisode` return an empty collection both for "already queued/pending" (normal, harmless) and genuine failures. The UI maps any empty result to a red "Could not queue cache: Unknown error" notification, so clicking Cache Now twice looks like a bug when it isn't.

7. **Stale cancellation flag can stall a legitimate re-dispatch** (`app/Models/CachedContentFile.php:178`, `DownloadCachedContentFile.php:162`)
   The pending-cancellation cache key is scoped by `content_fingerprint`, not row id, and lives for 10 minutes. Cancel a Pending download, then immediately re-request the same content: the new job reads the old flag and silently no-ops, leaving the new row stuck at `Pending`.

8. **Retention `deleteIds()` ignores a `never_expire` race** (`app/Services/CachedContentRetentionService.php:227`)
   Re-checks `status` before deleting (per its own docblock, "the only race window") but not `never_expire`. A row flagged `never_expire=true` between selection and deletion still gets deleted.

9. **"Read-only" DynamicGroups tab still allows destructive actions** (`app/Filament/Resources/DynamicGroups/RelationManagers/ChannelsRelationManager.php:116`)
   The new action-stripping code (meant to keep this tab strictly read-only) omits `delete`, `manual_tmdb_search`, and `process_vod` from its exclude list. Filament's `mountAction()` only checks `isDisabled()`/`authorize()`, not `visible()`, so these remain callable via a direct `mountTableAction()` call and can delete the underlying VOD channel.

10. **Disabled dynamic groups keep serving stale content** (`app/Jobs/SyncDynamicGroups.php:179`, `app/Services/XtreamCategoryService.php:221`)
    Stale rules used to be hard-deleted (cascading membership cleanup via FK). Now they're soft-disabled (`enabled=false`) with no cleanup of `dynamic_group_items`, and `applyDynamicGroupFilter()` never checks `enabled`. A renamed/removed rule's category keeps serving its frozen last-known member list indefinitely, independent of the caching feature.

### Correctness - lower priority

- **Suffix byte-range requests unsupported** (`app/Http/Controllers/Concerns/StreamLocalFile.php:42`): `RANGE_PATTERN` requires a digit before the dash, so `Range: bytes=-65536` (common for probing trailing metadata) silently falls back to a full 200 response, streaming the entire file.
- **TOCTOU on `filesize()`** (`app/Http/Controllers/CachedContentStreamController.php:118`): if the file disappears between `Storage::exists()` and `filesize()`, the `false` return fails the strictly-typed `int $fileSize` param and throws an uncaught `TypeError` (500) instead of a 404.

### Migration / schema safety

- **Missing `CONCURRENTLY` on indexes touching hot tables** (`database/migrations/2026_09_16_160100_add_user_id_index_to_cached_content_file_dynamic_groups.php:30`, and the similar `user_id`/`playlist_id` index migrations on `cached_content_files`): plain `Blueprint::index()` takes a table-level lock on Postgres for the build duration. These tables are written to continuously by `SyncDynamicGroups` and `DownloadCachedContentFile`; running this during a deploy risks stalling those job chains.

### Reuse / duplication

- **Download job bypasses the existing provider-throttling trait** (`app/Jobs/DownloadCachedContentFile.php:155`): reimplements retry/backoff instead of using `App\Traits\ProviderRequestDelay::withProviderThrottling()`, which every other provider-fetching job in the app uses for concurrency limiting and operator-configured delay. A `cache:content` sweep can fire unlimited concurrent multi-GB downloads at a provider with none of the app's usual throttling, risking provider bans.
- **Playlist auth-resolution logic duplicated a third time** (`app/Http/Controllers/CachedContentStreamController.php:56` vs `DvrStreamController::resolveUser()` vs `XtreamStreamController`): a security-relevant path now needs any future fix applied in three places.
- **Byte-range streaming logic duplicated** rather than shared with `DvrStreamController::stream()`, despite `StreamLocalFile`'s own docblock claiming both controllers use it.
- **`resolveChannel()`/`resolveEpisode()`** in `CachedContentActivityWidget.php:723` are ~25-line near-identical methods differing only by model class.
- **Retention `evaluate()` and `evaluateForDynamicGroup()`** (`CachedContentRetentionService.php:51`) duplicate the same "live-fingerprint set, stream candidates, collect stale ids" algorithm; the two paths have already drifted on empty-set handling.
- **`isCrossPlaylistDuplicate()`** (`CachedContentDispatchService.php:184`) is a dead wrapper with no production callers besides its own test.

### Efficiency

- `dispatchNew()` constructs two `DownloadCachedContentFile` job instances per dispatch, one is discarded (`CachedContentDispatchService.php:264`).
- Widget `poll('5s')` re-runs expensive title-resolution subqueries for every row on every tick, even static Completed/Failed rows (`CachedContentActivityWidget.php:98`).

### Conventions

- Four `TextColumn::placeholder('-')` calls in `CachedContentActivityWidget.php:144` use a literal em-dash, violating the workspace's no-em-dash rule for new code.

---

**Suggested priority for the author:** items 1-10 under Correctness affect real user-facing behavior (wrong content served, stuck jobs, a setting that silently does nothing, a bypassable "read-only" surface) and should be addressed before merge. The migration/schema-safety item is a deploy-time risk worth a quick fix (`->algorithm('concurrently')` or a raw statement). The reuse/efficiency/conventions items are good follow-up cleanup but not blocking.

Grimothy:
> @Grimothy
> 
> ## Review notes for #1524 (cache download feature)
> Automated multi-angle review across correctness, security, reuse, efficiency, and conventions. Grouped by severity, file:line references are against `pr-1524`.
> 
> ### Correctness - high priority
> 1. **Cross-playlist cache sharing is dispatched but never served** (`app/Services/CachedContentDispatchService.php:232`, `app/Http/Controllers/XtreamStreamController.php:571`)
>    `dispatchNew()` skips creating a new cache row whenever _any_ playlist already has that fingerprint, with no `playlist_id` or `share_cache_across_playlists` check. But `resolveCacheHit()` only serves `Completed` rows scoped to the requesting playlist's own `playlist_id`. Net effect: playlist B silently never gets its own cached copy when playlist A already cached the same fingerprint, and even when `share_cache_across_playlists` is intentionally enabled, the shared file is never actually served at playback time.
> 2. **Cancelled downloads get stuck at "Downloading" forever** (`app/Jobs/DownloadCachedContentFile.php:179`, `:239`)
>    Both cancellation checks return from `handle()` without setting `status` to `Failed`. Horizon reports the job as successful and never retries, and `atomicReclaim()` only reclaims `Pending`/`Failed` rows, so the row is a permanent zombie requiring a manual DB edit.
> 3. **Fingerprint collision for unmatched content** (`app/Models/CachedContentFile.php:128`, `Channel.php:217`, `Episode.php:115`)
>    When `tmdb_id`/`tvdb_id` are both null, the fingerprint collapses to a constant string (e.g. `"movie::::"`). Two different unmatched movies on the same playlist can collide, causing the wrong cached file to be served for the second one.
> 4. **New `cache_retention_mode` playlist setting is never read** (`app/Services/CachedContentRetentionService.php`)
>    The PR adds this column and wires it into the Playlist form, but the retention service never imports `Playlist` or checks the field, only `never_expire` and live-fingerprint membership. Setting a playlist to `manual`/`never-expire` retention mode is a silent no-op, files still get swept.
> 5. **`enable_cache` kill switch not enforced everywhere** (`app/Console/Commands/CacheContentCommand.php`, `CachedContentDispatchService.php`, `DownloadCachedContentFile.php`)
>    Only checked in UI action visibility and the two serve controllers. Running `php artisan cache:content --playlist=42` (the documented ad-hoc invocation) while "Enable Cache" is off in Settings still dispatches real downloads.
> 6. **"Cache Now" shows a false failure notification** (`app/Filament/Resources/Vods/VodResource.php:2540`, `EpisodesRelationManager.php:492`)
>    `dispatchForChannel`/`dispatchForEpisode` return an empty collection both for "already queued/pending" (normal, harmless) and genuine failures. The UI maps any empty result to a red "Could not queue cache: Unknown error" notification, so clicking Cache Now twice looks like a bug when it isn't.
> 7. **Stale cancellation flag can stall a legitimate re-dispatch** (`app/Models/CachedContentFile.php:178`, `DownloadCachedContentFile.php:162`)
>    The pending-cancellation cache key is scoped by `content_fingerprint`, not row id, and lives for 10 minutes. Cancel a Pending download, then immediately re-request the same content: the new job reads the old flag and silently no-ops, leaving the new row stuck at `Pending`.
> 8. **Retention `deleteIds()` ignores a `never_expire` race** (`app/Services/CachedContentRetentionService.php:227`)
>    Re-checks `status` before deleting (per its own docblock, "the only race window") but not `never_expire`. A row flagged `never_expire=true` between selection and deletion still gets deleted.
> 9. **"Read-only" DynamicGroups tab still allows destructive actions** (`app/Filament/Resources/DynamicGroups/RelationManagers/ChannelsRelationManager.php:116`)
>    The new action-stripping code (meant to keep this tab strictly read-only) omits `delete`, `manual_tmdb_search`, and `process_vod` from its exclude list. Filament's `mountAction()` only checks `isDisabled()`/`authorize()`, not `visible()`, so these remain callable via a direct `mountTableAction()` call and can delete the underlying VOD channel.
> 10. **Disabled dynamic groups keep serving stale content** (`app/Jobs/SyncDynamicGroups.php:179`, `app/Services/XtreamCategoryService.php:221`)
>     Stale rules used to be hard-deleted (cascading membership cleanup via FK). Now they're soft-disabled (`enabled=false`) with no cleanup of `dynamic_group_items`, and `applyDynamicGroupFilter()` never checks `enabled`. A renamed/removed rule's category keeps serving its frozen last-known member list indefinitely, independent of the caching feature.
> 
> ### Correctness - lower priority
> * **Suffix byte-range requests unsupported** (`app/Http/Controllers/Concerns/StreamLocalFile.php:42`): `RANGE_PATTERN` requires a digit before the dash, so `Range: bytes=-65536` (common for probing trailing metadata) silently falls back to a full 200 response, streaming the entire file.
> * **TOCTOU on `filesize()`** (`app/Http/Controllers/CachedContentStreamController.php:118`): if the file disappears between `Storage::exists()` and `filesize()`, the `false` return fails the strictly-typed `int $fileSize` param and throws an uncaught `TypeError` (500) instead of a 404.
> 
> ### Migration / schema safety
> * **Missing `CONCURRENTLY` on indexes touching hot tables** (`database/migrations/2026_09_16_160100_add_user_id_index_to_cached_content_file_dynamic_groups.php:30`, and the similar `user_id`/`playlist_id` index migrations on `cached_content_files`): plain `Blueprint::index()` takes a table-level lock on Postgres for the build duration. These tables are written to continuously by `SyncDynamicGroups` and `DownloadCachedContentFile`; running this during a deploy risks stalling those job chains.
> 
> ### Reuse / duplication
> * **Download job bypasses the existing provider-throttling trait** (`app/Jobs/DownloadCachedContentFile.php:155`): reimplements retry/backoff instead of using `App\Traits\ProviderRequestDelay::withProviderThrottling()`, which every other provider-fetching job in the app uses for concurrency limiting and operator-configured delay. A `cache:content` sweep can fire unlimited concurrent multi-GB downloads at a provider with none of the app's usual throttling, risking provider bans.
> * **Playlist auth-resolution logic duplicated a third time** (`app/Http/Controllers/CachedContentStreamController.php:56` vs `DvrStreamController::resolveUser()` vs `XtreamStreamController`): a security-relevant path now needs any future fix applied in three places.
> * **Byte-range streaming logic duplicated** rather than shared with `DvrStreamController::stream()`, despite `StreamLocalFile`'s own docblock claiming both controllers use it.
> * **`resolveChannel()`/`resolveEpisode()`** in `CachedContentActivityWidget.php:723` are ~25-line near-identical methods differing only by model class.
> * **Retention `evaluate()` and `evaluateForDynamicGroup()`** (`CachedContentRetentionService.php:51`) duplicate the same "live-fingerprint set, stream candidates, collect stale ids" algorithm; the two paths have already drifted on empty-set handling.
> * **`isCrossPlaylistDuplicate()`** (`CachedContentDispatchService.php:184`) is a dead wrapper with no production callers besides its own test.
> 
> ### Efficiency
> * `dispatchNew()` constructs two `DownloadCachedContentFile` job instances per dispatch, one is discarded (`CachedContentDispatchService.php:264`).
> * Widget `poll('5s')` re-runs expensive title-resolution subqueries for every row on every tick, even static Completed/Failed rows (`CachedContentActivityWidget.php:98`).
> 
> ### Conventions
> * Four `TextColumn::placeholder('-')` calls in `CachedContentActivityWidget.php:144` use a literal em-dash, violating the workspace's no-em-dash rule for new code.
> 
> **Suggested priority for the author:** items 1-10 under Correctness affect real user-facing behavior (wrong content served, stuck jobs, a setting that silently does nothing, a bypassable "read-only" surface) and should be addressed before merge. The migration/schema-safety item is a deploy-time risk worth a quick fix (`->algorithm('concurrently')` or a raw statement). The reuse/efficiency/conventions items are good follow-up cleanup but not blocking.

This has been the biggest pain :D

Grimothy:
@sparkison  - Picking this back up

Grimothy:
@sparkison Thanks for the detailed review. I've worked through it and rebased onto the latest `dev`. Here's what changed, plus a few spots where I went a different way and why.

## The big ones

**1. Sharing cache between playlists**
This turned out to be worse than it looked. The fingerprint was unique across the whole table, so once anyone cached a movie, no other playlist could get its own copy. That included other users' playlists.
- Each playlist now gets its own copy.
- Sharing only happens between playlists owned by the **same user**, and only when the playlist that has the file has "Share cache" turned on.
- Playback, the Cache Now button, the "is cached" badge and the stream endpoint all use the same rule now, so they can't disagree.
- This ships as a new migration, so anyone who already ran the old ones picks it up.

**2. Cancelled downloads stuck on "Downloading"**
Cancelling actually deletes the row, so this could only happen in a narrow race. Even so, every cancel path now marks a surviving row as Failed ("Cancelled") instead of leaving it stuck.

**3. Unmatched movies colliding**
Items with no TMDB/TVDB id now get their own unique key, so two unmatched movies no longer share one cached file.

**4. Retention mode did nothing**
You were right, and the global setting wasn't used either. The per-playlist setting now wins, with the global one as the fallback. "Never expire" and "Manual" both stop the automatic cleanup. "Time-based" never had an age limit behind it, so I renamed it to "Automatic (remove when content leaves the playlist)", which is what it really does.

**5. Enable Cache switch**
Turning caching off now stops everything:
- the `cache:content` command
- every dispatch path
- jobs that were already queued (they're marked "Caching disabled")

**6. Cache Now showing a red error**
Clicking it on something that's already cached or queued now shows "Already cached" or "Already queued for caching" instead of an error.

**7. Stale cancel flag**
The real problem: when a queued download was cancelled, its row was deleted before the job ever read the flag. The flag then hung around and blocked the *next* download of the same title. I removed that flag completely. Deleting the row already covers that case.

**8. Never-expire race**
The delete step now re-checks "never expire" right before deleting.

**9. Read-only Dynamic Groups tab**
Instead of hiding a list of actions, the tab now only allows the two cache actions. Everything else is removed, so actions like delete can't be triggered at all.

**10. Disabled groups still serving content**
When a rule is removed, its member list is now cleared. Disabled groups are also filtered out of the category lookup. If the rule is added back later, it's re-enabled and repopulated.

## Smaller fixes

- **Range requests:** `bytes=-N` (the last N bytes of a file) now works. Players use this to read metadata at the end of a file.
- **Vanishing files:** if a file disappears mid-request, you get a 404 instead of a 500.
- **DVR streaming** now uses the same range-handling code as the cache stream.
- **Login lookup:** the cache and DVR stream controllers now share one class for resolving playlist logins.
- **Widget:**
  - stores the title when the download starts, so polling doesn't re-run lookups for every row
  - refreshes every 5s while something is downloading, and every 60s otherwise
- **Other cleanup:**
  - merged the duplicate channel/episode lookup methods
  - removed the unused `isCrossPlaylistDuplicate()`
  - stopped creating the download job twice
  - replaced the em-dashes
- **"Share cache by default":** this setting was never used either. It now pre-fills the toggle when you create a playlist.

## Where I went a different way

- **Provider throttling:** I didn't wrap the whole download in `withProviderThrottling()`. Its slots are shared with imports and expire after 5 minutes, and a movie download can take an hour, so a few downloads could block M3U/EPG imports. Throttling now covers just opening the connection. The number of downloads running at once is still limited by the `cache-queue` worker count.
- **`CONCURRENTLY` on indexes:** both cache tables are brand new in this PR, so they're empty at deploy and nothing else writes to them yet. Nothing gets locked. The `playlists` changes are plain nullable columns, which don't lock anything meaningful on Postgres.
- **Xtream login lookup:** I left `XtreamStreamController` on its own code. It also accepts alias username/password logins, which the other two don't. Moving it over would change how logins work on the main playback path. Happy to do that as a separate change if you want it.

## One thing I left as-is

If a playlist or group briefly has no members, for example because cleanup runs in the middle of a sync, cleanup still removes all of its unprotected cached files. That's how it worked before, so I didn't change it here. I can add a guard if you'd like.

## Testing

- All the cache, DVR stream, Xtream category and dynamic-group tests pass: 325 in the files this PR touches, plus about 420 more across other tests that use the changed classes.
- The security scan is clean (composer audit, Vet, npm audit, checkpoint).
- Pint and the lang merge have both been run.

sparkison:
> This has been the biggest pain :D

haha! It's a 12k line monster!! 🤣

Grimothy:
Quick follow-up. I've added the tests that were still missing, and they turned up two things worth fixing:

**Orphan cleanup could delete a running download.** `cache:cleanup-orphans` deleted any row with no file that was *created* more than N days ago, no matter its status. So if a download failed a week ago and was retried today, cleanup could delete its row while it was still downloading. The finished file would then sit on disk with nothing pointing to it. Now it never touches rows that are currently downloading, and it goes by when the row was last updated instead of when it was created.

**The SSRF guard now blocks multicast addresses** (224.0.0.0/4 and ff00::/8). I left the 100.64.0.0/10 range allowed on purpose, since some people reach their provider over Tailscale or a carrier NAT.

**New tests:**
- `PrivateNetworkGuard`: private and reserved ranges, IPv6, blocked URL schemes, `user@host` tricks
- the download redirect handling: safe chains, redirects to private or non-web addresses, relative redirects, too many redirects
- the orphan cleanup command

94 tests pass across those three files, and the security scan is clean.

One trade-off: if a worker crashes in the middle of a download, that row now stays on "Downloading" until someone removes it by hand. That's the price of never deleting a download that's actually running. I've listed it under Known Gaps in the description.

